PR #68, opened 2020-10-24
Fix the file name of bash-completion

### LINKED ISSUE #35: fix source bash_completion in readme
**Note:** this issue was imported from https://github.com/wercker/stern/pull/116, but it was originally created by [anhvngo](https://github.com/anhvngo)...

fixed when we source `source <(brew --prefix)/etc/bash-completion` 
`bash: /dev/fd/63/etc/bash-completion` also the file name is wrong. It is not dash but underscore

superbrothers:
Fixes #35

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ Note that bash-completion must be sourced before sourcing the stern bash
 completion code in `.bashrc`.
 
 ```sh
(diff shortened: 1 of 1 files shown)


floryut (approved):
Looks good 👍

PR #69, opened 2020-10-24
Add information about bash-completion@2

superbrothers:
If running Bash 4.1+, bash-completion@2 should be used instead of
bash-completion.


floryut (approved):
💯

Issue #41, opened 2020-10-22
Stern does not support env variable settings defined in kube config


superbrothers:
**Note:** this issue was imported from https://github.com/wercker/stern/issues/122, but it was originally created by [michaelgeorgeattard](https://github.com/michaelgeorgeattard)...

Given the following valid kube config section used to connect to an EKS cluster:

```
users:
- name: xxx
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      args:
      - token
      - -i
      - xxx
      command: aws-iam-authenticator
      env:
      - name: AWS_PROFILE
        value: foo
```

Stern or its dependencies are not switching the `AWS_PROFILE` to `foo`, thus failing to connect to the EKS cluster.

PR #70, opened 2020-10-24
Load all kubernetes auth plugins

superbrothers:
It make all auth plugins available, including exec and openstack
plugins.

Closes https://github.com/stern/stern/issues/41


floryut (approved):
🥇

Issue #55, opened 2020-10-22
Installation error: (type *"k8s.io/client-go/kubernetes".Clientset has no field or method Core)


superbrothers:
**Note:** this issue was imported from https://github.com/wercker/stern/issues/142, but it was originally created by [fadhilkurnia](https://github.com/fadhilkurnia)...

Hi, I want to install stern using go get but get these errors. It will be better if stern use `go.mod` and `go.sum` since govendor is already deprecated. Thanks!

```
$ go get github.com/wercker/stern 

# github.com/wercker/stern/stern
src/github.com/wercker/stern/stern/main.go:46:45: clientset.Core undefined (type *"k8s.io/client-go/kubernetes".Clientset has no field or method Core)
src/github.com/wercker/stern/stern/main.go:70:29: clientset.Core undefined (type *"k8s.io/client-go/kubernetes".Clientset has no field or method Core)
src/github.com/wercker/stern/stern/tail.go:106:28: not enough arguments in call to req.Stream
	have ()
	want (context.Context)
src/github.com/wercker/stern/stern/watch.go:47:25: not enough arguments in call to i.Watch
	have ("k8s.io/apimachinery/pkg/apis/meta/v1".ListOptions)
	want (context.Context, "k8s.io/apimachinery/pkg/apis/meta/v1".ListOptions)

```

PR #71, opened 2020-10-24
Change the installation method by "build from source" to go get

superbrothers:
We have already changed to use go modules so we can install with `go get`.

I tested it in the following way:
```
$ docker run --rm golang:1.15 sh -c "go get -u github.com/stern/stern && bin/stern --version"
Unable to find image 'golang:1.15' locally
1.15: Pulling from library/golang
Digest: sha256:1ba0da74b20aad52b091877b0e0ece503c563f39e37aa6b0e46777c4d820a2ae
Status: Downloaded newer image for golang:1.15
stern version 1.11.0
```

Closes https://github.com/stern/stern/issues/55

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,10 +6,6 @@ GO ?= GO111MODULE=on GOPROXY=https://gocenter.io go
 build:
 	$(GO) build -o dist/stern .
 
-.PHONY: install
-install:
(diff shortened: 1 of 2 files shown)


floryut (approved):
Easier indeed, good job

PR #74, opened 2020-10-25
Prevent stern completion functions from redefining kubectl's them

### LINKED ISSUE #50: Stern bash completions clash with kubectl completions
**Note:** this issue was imported from https://github.com/wercker/stern/issues/137, but it was originally created by [drzero42](https://github.com/drzero42)...

I just spent some time troubleshooting a weird problem I was experiencing with kubectl on my system. I noticed that recently I could not get kubectl to autocomplete names of deployments, statefulsets, replicasets and pods. It would gladly autocomplete namespaces, crds, storageclasses, secrets, configmaps and others. As most of my cluster don't have anything in the default namespace, but rather put everything into other namespaces, at first I thought name completion was totally broken, but it turned out that it was just looking in the default namespace, even though I was trying to tab-complete the name of a pod in a different namespace, eg: `kubectl -n kube-system describe pod <tab><tab>`.

I load bash completions from my .bashrc with `source <(kubectl completion bash)`. I also load completions for a bunch of other commands, like minikube, kind, argocd and also stern. I decided to try loading the kubectl completions by hand in my terminal, which made it work. So I went to my .bashrc and started experimenting with the order that I loaded the completions in, and quickly discovered that if stern completions are loaded after kubectl completions, then I can't autocomplete names from any other namespace than default.

I am in no way an expert on how to write bash completion functions, but a quick look at the output of `stern --completion bash` and comparing to the output of `kubectl completion bash`, I see stern defining a number of functions that kubectl also defines. Thus, if the stern bash completions are loaded after the kubectl completions, it screws them up. It would be great if the stern bash completions could work without fiddling with kubectl completions ;)

superbrothers:
By adding `__stern_` prefix to stern bash completion functions, we avoid
duplicating the name with kubectl bash completion functions.

Fixes #50

### DIFF
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -31,17 +31,17 @@ if ! which kubectl >/dev/null 2>&1; then
 	__is_kubectl_installed=false
 fi
 
-__kubectl_override_flag_list=(kubeconfig context namespace)
-__kubectl_override_flags()
+__stern_kubectl_override_flag_list=(kubeconfig context namespace)
+__stern_kubectl_override_flags()
 {
-    local ${__kubectl_override_flag_list[*]} two_word_of of
+    local ${__stern_kubectl_override_flag_list[*]} two_word_of of
     for w in "${words[@]}"; do
         if [ -n "${two_word_of}" ]; then
             eval "${two_word_of}=\"--${two_word_of}=\${w}\""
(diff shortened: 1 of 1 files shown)


floryut (approved):
🚀

PR #79, opened 2020-10-26
Use --skip-validate with goreleaser release at release

superbrothers:
This is to avoid the "git is currently in a dirty state" error for hack/tools/go.sum file.

- https://github.com/stern/stern/runs/1308798250?check_suite_focus=true#step:6:84

I didn't know how to get around this error. Tools are only used to build stern and do not affect the release.


floryut (approved):
Good, I indeed saw that

PR #82, opened 2020-10-27
Update version to 1.12.1

floryut:
We forgot to update the cli version, let's push a new release to fix that ?


superbrothers (approved):
Nice catch! :+1: I plan to modify stern to use git tag for the version!

floryut:
> Nice catch! 👍 I plan to modify stern to use git tag for the version!

Indeed could be cleaner!

PR #84, opened 2020-10-27
Use ldflags to set version information

superbrothers:
Version information is output in the following format:
```
$ stern --version
version: v1.12.1
commit: 2d4bd0de2d950b189eba7a7ebf8120283b7009c0
built at: 2020-10-27T14:13:08Z
```


floryut (approved):
🍾

PR #85, opened 2020-10-28
Run `go fmt` and `go vet` against code

superbrothers:
It keeps code to be easier to write, read, and maintain.


floryut (approved):
Neat ! 🏆

PR #94, opened 2020-11-05
Add a new installation way with asdf tool

superbrothers:
This PR adds a new installation way with asdf tool.

## REF

- https://github.com/looztra/asdf-stern/pull/3


floryut (approved):
👍

PR #99, opened 2020-11-15
README: Recommend `brew install stern` from homebrew-core

issyl0:
- As of https://github.com/Homebrew/homebrew-core/commit/7d44ae9f44d112cde586258a28b8bbed22b43f2a, this fork of `stern`, with its latest version, ships in Homebrew's core tap (on macOS and shortly on Linux).

PR #98, opened 2020-11-10
Update a Homebrew formula automatically on release

superbrothers:
I created the formula repository (https://github.com/stern/homebrew-stern) prior to this change.

Closes #97

PR #100, opened 2020-11-16
Revert "Update a Homebrew formula automatically on release (#98)"

superbrothers:
We are finally able to install this forked stern from homebrew-core. 🙌

```
$ brew install stern
```

## REF

- #99


floryut (approved):
That's awesome!

PR #102, opened 2020-11-21
Follow logs of new containers on restarting

superbrothers:
With this change, stern continues to follows logs of new containers on restarting.

Fixes #6, #7, #13 

You can check that issues are fixed with the following steps:

## Case 1

**Preparing**
```sh
cat <<'EOL' | kubectl apply -f-
apiVersion: v1
kind: Pod
metadata:
  name: foo
spec:
  containers:
    - name: foo
      image: busybox
      command: ["sh", "-c", "touch /log; tail -f /log"]
      readinessProbe:
        initialDelaySeconds: 5
        timeoutSeconds: 5
        exec:
          command: ["sh", "-c", "echo $(date) ::: Readiness probe >> /log; grep imready /foo"]
      livenessProbe:
        initialDelaySeconds: 5
        timeoutSeconds: 5
        exec:
          command: ["sh", "-c", "echo $(date) ::: Liveness probe >> /log; grep imalive /foo"]
EOL
```

**Before**
```sh
$ stern --version
version: 1.13.0
commit: 48835a04b83be525ee1125d308b025dabd741190
built at: 2020-11-06T07:46:45Z
$ stern foo
+ foo › foo
foo foo Sat Nov 21 03:53:31 UTC 2020 ::: Readiness probe
foo foo Sat Nov 21 03:53:31 UTC 2020 ::: Liveness probe
foo foo Sat Nov 21 03:53:41 UTC 2020 ::: Readiness probe
foo foo Sat Nov 21 03:53:41 UTC 2020 ::: Liveness probe
foo foo Sat Nov 21 03:53:51 UTC 2020 ::: Readiness probe
foo foo Sat Nov 21 03:53:51 UTC 2020 ::: Liveness probe
foo foo Sat Nov 21 03:54:01 UTC 2020 ::: Readiness probe
foo foo Sat Nov 21 03:54:11 UTC 2020 ::: Readiness probe
foo foo
# <---- Do not follow logs of the restarted container
```

**After**
```sh
$ dist/stern foo
+ foo › foo
foo foo Sat Nov 21 03:52:31 UTC 2020 ::: Readiness probe
foo foo Sat Nov 21 03:52:31 UTC 2020 ::: Liveness probe
foo foo Sat Nov 21 03:52:41 UTC 2020 ::: Readiness probe
foo foo Sat Nov 21 03:52:41 UTC 2020 ::: Liveness probe
foo foo Sat Nov 21 03:52:51 UTC 2020 ::: Readiness probe
[...]

PR #103, opened 2020-11-22
Make stern goroutine safe

### LINKED ISSUE #4: stern not thread safe
**Note:** this issue was imported from https://github.com/wercker/stern/issues/15, but it was originally created by [JrCs](https://github.com/JrCs)...

Hi,
i think stern is not thread safe, as you use 2 go functions that use the same map (tails) without using some synchronisation like mutex.

superbrothers:
Fixes #4 

This PR is based on #102, so please review it first 🍻

### DIFF
--- a/stern/stern.go
+++ b/stern/stern.go
@@ -16,10 +16,33 @@ package stern
 
 import (
 	"context"
+	"sync"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


floryut (approved):
Nice, lgtm

PR #106, opened 2020-12-03
Fix mixed log lines

### LINKED ISSUE #104: The container information prefix is output on different log lines
**What happened**:

image

The container information prefix is output on correct log lines.

**Environment**:
- stern version (use `stern --version`):
    - version: 1.13.0
    - commit: 48835a04b83be525ee1125d308b025dabd741190
    - built at: 2020-11-06T07:46:45Z
- OS (e.g: `cat /etc/os-release`):
- Install tools (e.g: Homebrew): asdf
- Others:

superbrothers:
> A template may be executed safely in parallel, although if parallel
executions share a Writer the output may be interleaved.

- https://golang.org/pkg/text/template/#Template.Execute

For the above reason, the result of the template's execution must be
stored in a buffer once.

Fixed #104

### DIFF
--- a/stern/tail.go
+++ b/stern/tail.go
@@ -16,6 +16,7 @@ package stern
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"fmt"
 	"hash/fnv"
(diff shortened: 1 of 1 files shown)


floryut (approved):
👍

PR #111, opened 2021-01-27
Add field selector

uesyn:
Add field selector similar to `kubectl get --field-selector`.
This PR make log filter more flexible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ stern frontend --selector release=canary
 
 Tail the pods on `kind-control-plane` node accross all namespaces
 ```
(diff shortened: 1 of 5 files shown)


superbrothers (approved):
Nice! 🍺

superbrothers (commented):
Could you add a new flag to the flag list in README.md? In addition, it would be nice if you could add an example to the Examples section.

uesyn:
@superbrothers 
Added this flag into flag list and example!

uesyn:
I missed to fix argument validation.
Fixed it.

superbrothers (approved):
:100:

PR #115, opened 2021-02-20
Use Go 1.16

superbrothers:
In addition, by updating goreleaser, stern provides additional support
for `darwin/arm64` platform.


floryut (approved):
Nice, I saw the release but didn't took the time!

PR #116, opened 2021-03-28
Use GITHUB_TOKEN to authenticate with ghcr.io

superbrothers:
ghcr.io now supports GITHUB_TOKEN.

https://github.blog/changelog/2021-03-24-packages-container-registry-now-supports-github_token/


floryut (approved):
That's indeed cleaner :)

PR #118, opened 2021-04-18
Add --timezone option to set timestamps to specific timezone

### LINKED ISSUE #43: Suggestion Time in Custom Format
**Note:** this issue was imported from https://github.com/wercker/stern/issues/125, but it was originally created by [derekrprice](https://github.com/derekrprice)...

I'd like to be able to add a short timestamp (no decimal seconds) in a custom format string.  My use case is mostly making the log lines fit better on my monitor.

superbrothers:
`--timezone` option sets timestamps to a specific time zone. The default
timezone is changed to your locale from UTC, so this is a breaking
change.

Print timestamps with your local timezone:
```
stern . --timestamps
```

Print timestamps with specific timezone:
```
stern . -t --timezone Europe/Paris
```

`TZ` environment also can be used:
```
TZ=Europe/Paris stern . -t
```

Closes #43


floryut (approved):
Great man 👍

Issue #117, opened 2021-04-05
Panic after watching pod for 30 min


dramich:
**What happened**:
Letting the command `stern -n cattle-system -l app=rancher --tail 100` run for 30 min will always panic

```
panic: interface conversion: runtime.Object is *v1.Status, not *v1.Pod

goroutine 11 [running]:
github.com/stern/stern/stern.Watch.func1(0x2272960, 0xc0000be1c0, 0xc0003ae500, 0x0, 0xc0003ae640, 0x0, 0x20e1090, 0x7, 0xc0007181e0, 0xc000718240, ...)
	/home/runner/work/stern/stern/stern/watch.go:66 +0xd9a
created by github.com/stern/stern/stern.Watch
	/home/runner/work/stern/stern/stern/watch.go:57 +0x2e6
```

**What you expected to happen**:
Stern continues to display logs
**How to reproduce it (as minimally and precisely as possible)**:
It constantly reproduces for me watching a rancher pod
**Anything else we need to know?**:

**Environment**:
- stern version (use `stern --version`): version: 1.14.0
commit: c3deff1bf1da4a94c072d3cbb315dadbf4415342
built at: 2021-01-28T13:35:24Z
- OS (e.g: `cat /etc/os-release`): macOS Cataline 10.15.7
- Install tools (e.g: Homebrew) Downloaded the binary
- Others:

superbrothers:
Thanks for the report.  I will check it out.

superbrothers:
@dramich I see that there is a problem with the code, but I cannot reproduce this bug. Using minikube or kind, can you reproduce the bug?

dramich:
I attempted with kind and running rancher through a helm chart but can't repro. I saw this running a very large setup, not locally unfortunately.

dramich:
I found a way to repro, if you have stern watching a pod and change the deployment, it will cause the same panic.

superbrothers:
Umm... I could not repro. Can you please share all the steps from creating the cluster locally to stern panic? Please provide versions of all relevant tools as well.

k1ng440:
I have also received this error today. I was running stern on a pod that is hosted on GKE (200+ ms ping)


```
unexpected error: http2: client connection lost
unexpected error: http2: client connection lost
panic: interface conversion: runtime.Object is *v1.Status, not *v1.Pod

goroutine 25 [running]:
[...]

PR #121, opened 2021-04-26
Fix "interface conversion: runtime.Object is *v1.Status, not *v1.Pod"

superbrothers:
If type conversion to a Pod object fails, skip the process.

ref/ #117


floryut (approved):
Hope that this will do 👍

Issue #101, opened 2020-11-20
stern shows less detail compared to `kubectl logs`


danielhoherd:
**What happened**:

I noticed that some very important bits were left out of the stern logs compared to what shows up in `kubectl logs`.

**What you expected to happen**:

Logs output would be consistent with or show more info than `kubectl logs` does.

**How to reproduce it (as minimally and precisely as possible)**:

I'm not sure how to reproduce this other than just looking at random logs to see the behavior. Here is an example of how I observed it in my environment:

```
$ kubectl logs -n kube-system --all-containers calico-node-n27q5
time="2020-11-20T21:25:48Z" level=info msg="Running as a Kubernetes pod" source="install.go:140"
time="2020-11-20T21:25:49Z" level=info msg="Installed /host/opt/cni/bin/bandwidth"
time="2020-11-20T21:25:49Z" level=info msg="Installed /host/opt/cni/bin/calico"
time="2020-11-20T21:25:50Z" level=info msg="Installed /host/opt/cni/bin/calico-ipam"
time="2020-11-20T21:25:50Z" level=info msg="Installed /host/opt/cni/bin/flannel"
time="2020-11-20T21:25:50Z" level=info msg="Installed /host/opt/cni/bin/host-local"
time="2020-11-20T21:25:50Z" level=info msg="Installed /host/opt/cni/bin/install"
time="2020-11-20T21:25:50Z" level=info msg="Installed /host/opt/cni/bin/loopback"
time="2020-11-20T21:25:50Z" level=info msg="Installed /host/opt/cni/bin/portmap"
time="2020-11-20T21:25:50Z" level=info msg="Installed /host/opt/cni/bin/tuning"
time="2020-11-20T21:25:50Z" level=info msg="Wrote Calico CNI binaries to /host/opt/cni/bin\n"
time="2020-11-20T21:25:50Z" level=info msg="CNI plugin version: v3.16.3-gke.0\n"
time="2020-11-20T21:25:50Z" level=info msg="/host/secondary-bin-dir is not writeable, skipping"
time="2020-11-20T21:25:50Z" level=info msg="Using CNI config template from CNI_NETWORK_CONFIG environment variable." source="install.go:319"
time="2020-11-20T21:25:50Z" level=info msg="Created /host/etc/cni/net.d/10-calico.conflist"
time="2020-11-20T21:25:50Z" level=info msg="Done configuring CNI.  Sleep= false"
{
  "name": "k8s-pod-network",
[...]

PR #124, opened 2021-05-08
Enable --init-containers option

rkmathi:
resolves: #101 

I noticed `--init-containers` options is not parsed.
Until this fix, it was always false.


superbrothers (approved):
Good catch!

Issue #48, opened 2020-10-22
Support show multiple container state log


superbrothers:
**Note:** this issue was imported from https://github.com/wercker/stern/issues/131, but it was originally created by [hulucc](https://github.com/hulucc)...

As the title.

superbrothers:
Does this issue wants to say is that you want to specify multiple container states as follows? 🤔 

```
$ stern --container-state running,terminated .
```

floryut:
> Does this issue wants to say is that you want to specify multiple container states as follows? 🤔
> 
> ```
> $ stern --container-state running,terminated .
> ```

Yup I guess 👍

PR #126, opened 2021-05-09
Support multiple --container-state option

rkmathi:
resolves: #48 

I changed `--container-state` option receiving multiple values.


superbrothers (approved):
:100:

PR #129, opened 2021-05-13
Make init container flag message more clear

KeisukeYamashita:
## What

This flag include or exclude the init container and the default value is `true`.

It's hard to understand wether **is it going to get included or excluded** because the messages are confusing. Therefore, I propose to make it clear by removing the `or exclude` term.


superbrothers (approved):
This change clarifies the function of the flag :+1:

floryut:
Easier to understand indeed 👍

Issue #146, opened 2021-08-08
Error: lost watch connection


kettenbach-it:
**What happened**:
stern loses the connection to k8s after a seemingly random period of time when tracking a namespace. This happens even if none of the pods that stern is tracking are restarted (in fact, no pods are restarted at all). It shows:
```Error: lost watch connection```

**What you expected to happen**:
stern should hold the connection permanently. 

**How to reproduce it (as minimally and precisely as possible)**:
It happens with any calls of the kind:
```
stern -t -p -s8h  -n <namespace>
```

**Environment**:
- stern version (use `stern --version`): 
`version: 1.19.0`
- OS (e.g: `cat /etc/os-release`): 
`Mac OS Big Sur 11.4`
- Install tools (e.g: Homebrew):
```
brew info stern
stern: stable 1.19.0 (bottled), HEAD
```

PR #149, opened 2021-08-09
Make watcher gets restarted in case of recoverable errors

### LINKED ISSUE #47: Reconnect when receiving GOAWAY
**Note:** this issue was imported from https://github.com/wercker/stern/issues/130, but it was originally created by [grosser](https://github.com/grosser)...

when streaming logs and killing 1/3 api servers, stern spits out this:
```
ERROR: logging before flag.Parse: E0205 21:13:53.792712   21184 streamwatcher.go:109] Unable to decode an event from the watch stream: http2: server sent GOAWAY and closed the connection; LastStreamID=5, ErrCode=NO_ERROR, debug=""
```
(I think the api-server it was connected to died) ... but then it just sits there and does nothing even if new logs lines arrive ... it should retry watching instead (ideally with offset so it does not duplicate)

superbrothers:
This PR uses RetryWatcher to make watcher gets restarted in case of recoverable errors. RetryWatcher will make sure that in case the underlying watcher is closed (e.g. due to API timeout or etcd timeout) it will get restarted from the last point without the consumer even knowing about it.

- https://github.com/kubernetes/client-go/blob/v0.22.0/tools/watch/retrywatcher.go#L59-L64

Fixes #47, #117, #146

### DIFF
--- a/stern/watch.go
+++ b/stern/watch.go
@@ -27,6 +27,8 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/watch"
 	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
+	watchtools "k8s.io/client-go/tools/watch"
 )
 
 // Target is a target to watch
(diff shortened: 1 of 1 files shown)


rkmathi:
Looks good 👍

rkmathi (dismissed):

floryut (dismissed):
💯

PR #157, opened 2021-09-11
Use Go 1.17

rkmathi:
https://groups.google.com/g/golang-announce/c/dx9d7IOseHw

The vulnerability is found in versions below 1.16.8 or 1.17.1 so will be updated to the latest Go 1.17.1.


superbrothers (approved):
There is a big change in `go.mod`, but it was changed in Go 1.17, right? I didn't know that, so I was a little surprised.

> For the go command to correctly resolve transitive imports using the pruned module graph, the go.mod file for each module needs to include more detail about the transitive dependencies relevant to that module. If a module specifies go 1.17 or higher in its go.mod file, its go.mod file now contains an explicit require directive for every module that provides a transitively-imported package. (In previous versions, the go.mod file typically only included explicit requirements for directly-imported packages.)

- https://golang.org/doc/go1.17#go-command

rkmathi:
Yes, I didn't even know it!
I just run go mod tidy with Go 1.17.1.

Issue #22, opened 2020-10-22
Tailing logs with JSON output.


superbrothers:
**Note:** this issue was imported from https://github.com/wercker/stern/issues/97, but it was originally created by [davinchia](https://github.com/davinchia)...

Do the current stern options support JSON log output from pods with traditional stern color?

A subset of our pods currently output JSON logs with additional fields. An example Stern output (Stern 1.10):

> dsj-stat-builder-service-daemon-55444b74cb-zwtfx dsj-stat-builder-service-daemon {"@timestamp":"2019-01-08T03:13:41.573Z","source_host":"dsj-stat-builder-service-daemon-55444b74cb-zwtfx","file":"Bucket.java","method":"readMetadataFromFS","level":"INFO","line_number":"441","thread_name":"Step Runner for count-records-and-segments","@version":1,"logger_name":"com.rapleaf.formats.bucket.Bucket","message":"Metadata obtained at gs:\/\/liveramp-eng-dist-incoming-data\/input_data\/data\/spruce\/persistent\/data_sync_requests_by_job_id\/1610409_1546917039906\/bucket.meta","class":"com.rapleaf.formats.bucket.Bucket","mdc":{"run_identifier":"2010629208","joblet_identifier":"1736402519276071327217165","application":"dsj-stat-builder","workflow_execution_id":"344533283","workflow_attempt_id":"355297393","source_host":"dsj-stat-builder-service-daemon-55444b74cb-zwtfx","team":"dev-dist"}} 

I am able to parse these using JQ, but haven't been able to figure out how to preserve Stern's color coding while doing so. Is this currently supported/I need to read the docs better or is additional parsing of json log messages not yet a feature?

superbrothers:
**Note:** this comment was imported, but it was originally made by [hatchan](https://github.com/hatchan)...

When using the `output` flag `json` or `raw` (https://github.com/wercker/stern#templates) we do not add any colors. 

If you want to have a combination of `json`/`raw` and `default` you have to create your own custom template. You can see how the current `output` flags map to their templates here: https://github.com/wercker/stern/blob/807830e577193a63ede18bb8c7a8b86df8694c00/cmd/cli.go#L216-L236

superbrothers:
**Note:** this comment was imported, but it was originally made by [davinchia](https://github.com/davinchia)...

Thanks @hatchan for your quick reply.

[...]

PR #160, opened 2021-09-20
Add a new function `parseJSON` to --template flag

superbrothers:
`parseJSON` function parses string as JSON.

For example, if an application outputs the following log message:

```
{"level": "INFO", "message": "This is log message"}
```

We can use `parseJSON` function to format logs as follows:

```
$ stern . --template='{{.PodName}}/{{.ContainerName}} {{with $d := .Message | parseJSON}}[{{$d.level}}] {{$d.message}}{{end}}{{"\n"}}'
app-544bbcd5dc-fwgdj/application [INFO] This is log message
```

You can try this new function using the following steps:
```
cat <<EOL | kubectl apply -f-
apiVersion: apps/v1
kind: Deployment
metadata:
  name: app
spec:
  selector:
    matchLabels:
      app: app
  template:
    metadata:
      labels:
        app: app
    spec:
      containers:
      - image: busybox
        name: application
        command: ["/bin/sh"]
        args:
        - -c
        - |-
          while true; do
            echo '{"level": "INFO", "message": "This is log message"}'
            sleep 1
          done
EOL

stern . --template='{{.PodName}}/{{.ContainerName}} {{with $d := .Message | parseJSON}}[{{$d.level}}] {{$d.message}}{{end}}{{"\n"}}'
```

Closes https://github.com/stern/stern/issues/22

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,11 +114,11 @@ will receive the following struct:
 The following functions are available within the template (besides the [builtin
 functions](https://golang.org/pkg/text/template/#hdr-Functions)):
 
-| func    | arguments             | description                                                     |
-|---------|-----------------------|-----------------------------------------------------------------|
-| `json`  | `object`              | Marshal the object and output it as a json text                 |
-| `color` | `color.Color, string` | Wrap the text in color (.ContainerColor and .PodColor provided) |
-
+| func        | arguments             | description                                                     |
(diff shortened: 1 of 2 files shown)


rkmathi (approved):
Looks great!
this is very convenient😺

PR #162, opened 2021-10-05
Add and export follow option to TailOptions

arthurcgc:
Hi guys,
I'm using stern as a lib inside the project [rpaas-operator](https://github.com/tsuru/rpaas-operator/blob/master/internal/pkg/rpaas/log.go) to log the pods of the CRD we manage... But it is of value to us to have the option to not always tail the logs and just print out the last N lines of logs from the instance, hence I made this PR to add this flexibility. 

cc: @nettoclaudio

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -287,6 +287,7 @@ func (o *options) sternConfig() (*stern.Config, error) {
 		FieldSelector:         fieldSelector,
 		TailLines:             tailLines,
 		Template:              template,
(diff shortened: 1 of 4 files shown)


floryut:
@arthurcgc You should also add this new option to https://github.com/stern/stern/blob/master/README.md#cli-flags 👍

superbrothers:
This PR does not add any flags, so there should be no update to README.

Issue #167, opened 2021-11-25
Build from source fails


vhermecz:
I wanted to install stern by building from source
However, it did not succeed...

```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.3 LTS
Release:	20.04
Codename:	focal
$ snap install go --classic
...
$ go version
go version go1.17.3 linux/amd64
$ go get -u github.com/stern/stern
...
go get: installing executables with 'go get' in module mode is deprecated.
	Use 'go install pkg@version' instead.
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.
$ stern
stern: command not found
$ go install -u github.com/stern/stern
flag provided but not defined: -u
usage: go install [build flags] [packages]
Run 'go help install' for details.
$ go install github.com/stern/stern
go install: version is required when current directory is not in a module
	Try 'go install github.com/stern/stern@latest' to install the latest version
$ go install github.com/stern/stern@latest
go install: github.com/stern/stern@latest (in github.com/stern/stern@v1.20.1):
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
$ stern
stern: command not found
```

superbrothers:
I'll check it.

PR #168, opened 2021-12-01
Remove the pinned version of klog

superbrothers:
klog 2.30.0 has been released and the dependency issue with logr has been resolved. Therefore, we will remove the pinned version of logr.

Fixes https://github.com/stern/stern/issues/167


rkmathi (approved):
Looks good 😺

PR #169, opened 2021-12-01
Use "go install" instead of "go get"

superbrothers:
We can use `go install` command to install an executable binary since Go 1.16.


rkmathi (approved):
Looks good too 😸

PR #175, opened 2021-12-15
Disable uploading the krew plugin manifest using goreleaser

superbrothers:
I had set `skip_upload` to `false` when it should have been `true` 🙃


floryut (approved):
It happens 😄 
👍

PR #178, opened 2022-01-20
Added validation on datetime prefix message

### LINKED ISSUE #161: Logs not shown – `cannot parse "failed" as "2006"`
**What happened**:

```
unexpected error: parsing time "failed" as "2006-01-02T15:04:05.999999999Z07:00": cannot parse "failed" as "2006"
```

when following logs of a Job on k8s version `v1.19.13-eks-8df270`.

Logs should be shown.

Don't have a reproducible example. I'm not even sure whether this could be in stern, or is some kubernetes code issue. The error definitely hints at some Go date formatting code, and I only saw the issues with `stern`. I was able to see the Job being created, and could tail its logs with `k9s`.

**Environment**:
- stern version (use `stern --version`): 1.20.1
- OS (e.g: `cat /etc/os-release`): macOS 11.5
- Install tools (e.g: Homebrew): homebrew
- Others: Tried these versions of kubectl in PATH (if it even matters) – v1.22.2 and v1.19.10

odinnordico:
When `timestamp` flag is set to `true` but the cluster provide message without timestamp an error is shown instead of the actual message, so the user would not see what the server sent, this PR aims to not break the message but validate if there is a valid prefix date otherwise it will show the message prefixed with `[missing timestamp]`

It is a suggestion, just LMK if there is a better way or it just needs to be the way it is!

Fixes #161

### DIFF
--- a/stern/tail.go
+++ b/stern/tail.go
@@ -93,17 +93,13 @@ func (o TailOptions) UpdateTimezoneIfNeeded(message string) (string, error) {
 
 	idx := strings.IndexRune(message, ' ')
 	if idx == -1 {
-		return message, fmt.Errorf("missing timestamp")
+		return message, errors.New("missing timestamp")
 	}
 
 	datetime := message[:idx]
-	r, _ := regexp.Compile(`\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d(\.\d+([+-][0-2]\d:[0-5]\d|(Z)?))`)
-	if !r.MatchString(datetime) {
-		return message, fmt.Errorf("missing timestamp")
-	}
 	t, err := time.ParseInLocation(time.RFC3339Nano, datetime, time.UTC)
 	if err != nil {
-		return message, err
+		return message, errors.New("missing timestamp")
 	}
 
 	return t.In(o.Location).Format("2006-01-02T15:04:05.000000000Z07:00") + message[idx:], nil
--- a/stern/tail.go
+++ b/stern/tail.go
@@ -93,10 +93,14 @@ func (o TailOptions) UpdateTimezoneIfNeeded(message string) (string, error) {
 
 	idx := strings.IndexRune(message, ' ')
 	if idx == -1 {
-		return "", fmt.Errorf("a log message does not seem to have a datetime prefix: %s", message)
+		return "", fmt.Errorf("missing timestamp")
 	}
 
 	datetime := message[:idx]
+	r, _ := regexp.Compile(`\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d(\.\d+([+-][0-2]\d:[0-5]\d|(Z)?))`)
+	if !r.MatchString(datetime) {
+		return "", fmt.Errorf("missing timestamp")
+	}
 	t, err := time.ParseInLocation(time.RFC3339Nano, datetime, time.UTC)
 	if err != nil {
 		return "", err
@@ -212,12 +216,13 @@ func (t *Tail) ConsumeRequest(ctx context.Context, request rest.ResponseWrapper)
 				continue
 			}
 
-			msg, err := t.Options.UpdateTimezoneIfNeeded(msg)
+			updatedMsg, err := t.Options.UpdateTimezoneIfNeeded(msg)
 			if err != nil {
-				return err
+				t.Print(fmt.Sprintf("[%v] %s", err, msg))
+				continue
 			}
 
-			t.Print(msg)
+			t.Print(updatedMsg)
 		}
 
 		if err != nil {
--- a/stern/tail.go
+++ b/stern/tail.go
@@ -93,17 +93,17 @@ func (o TailOptions) UpdateTimezoneIfNeeded(message string) (string, error) {
 
 	idx := strings.IndexRune(message, ' ')
 	if idx == -1 {
-		return "", fmt.Errorf("missing timestamp")
+		return message, fmt.Errorf("missing timestamp")
 	}
 
 	datetime := message[:idx]
 	r, _ := regexp.Compile(`\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d(\.\d+([+-][0-2]\d:[0-5]\d|(Z)?))`)
 	if !r.MatchString(datetime) {
-		return "", fmt.Errorf("missing timestamp")
+		return message, fmt.Errorf("missing timestamp")
 	}
 	t, err := time.ParseInLocation(time.RFC3339Nano, datetime, time.UTC)
 	if err != nil {
-		return "", err
+		return message, err
 	}
 
 	return t.In(o.Location).Format("2006-01-02T15:04:05.000000000Z07:00") + message[idx:], nil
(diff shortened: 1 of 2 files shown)


superbrothers:
Is the `timestamp` flag enabled, but the log message returned by apiserver does not contain a timestamp? Do you know the steps that can reproduce the problem?

odinnordico:
It is when a `kubelet` throw the error so it is a bit hard to reproduce, i did it by using a [probes troublemaker](https://github.com/Praqma/k8s-probes-demo/blob/master/readiness-liveness-with-troublemaker.yaml), then i changes the `replicase` to 5 wait some seconds and finally deleting the deployment

here an screen of `stern troublemaker` with and without `-t` (check the rpc log message)

image


i used this yaml:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: readiness-with-troublemaker
  labels:
    app: readiness-with-troublemaker
    
spec:
  replicas: 5
  selector:
    matchLabels:
      app: readiness-with-troublemaker
  template:
    metadata:
      labels:
        app: readiness-with-troublemaker
    spec:
      containers:
      - name: troublemaker 
        image: kamranazeem/k8s-probes-demo:latest
        imagePullPolicy: Always

        # Notes: 
        # * ENTRYPOINT in Docker = 'command' in k8s
        # * CMD in Docker = 'args' in k8s 
        # * If you provide "command", then you also need to provide "args".
        #     For some reason the CMD defined in the container image is ignored,
        #     when k8s "command" is specified.
        command: ["/troublemaker.sh"]
        args: ["nginx", "-g", "daemon off;"]

        env:
        # Double quotes for values of ENV variables are important.
        - name: ROLE
          #value: "NORMAL"
          value: "TROUBLEMAKER"

        volumeMounts:
        - name: shared-directory
          mountPath: /shared
        
        ports:
        - name: alt-http
          containerPort: 8888

        resources:
          limits:
            cpu: 10m
            memory: 20Mi
          requests:
            cpu: 5m
            memory: 20Mi
      - name: liveness-with-troublemaker2 # Name of container in the pod.
        image: kamranazeem/k8s-probes-demo:latest

        env:
        - name: START_DELAY
          value: "60" # Double quotes for ENV variables are important.

        ports:
        - name: http 
          containerPort: 80

        resources:
          limits:
            cpu: 10m
            memory: 20Mi
          requests:
            cpu: 5m
            memory: 20Mi
       
        # Liveliness probes are used while the container is up, 
        #   to see if the service is "still" alive.
        # If the service does not respond, then kubernetes kills the container/pod,
        #   and restarts it.
        # It is therefore important to have initialDelaySeconds for the "livenessProbe"
        #   to be as late as possible.
        # If it is a small number, and if for some reason container is not started yet,
        #   (in case of slow application startup), then kubernetes will restart it.
        # This means the pod will stuck in a crash loop, and will never start.
        # Remember, livenessProbe does NOT start AFTER readiness probe.
        #   In-fact, they both start at the same time.
        # Liveness probe will vary from application to application. 
        #   Some start fast, some start slow.
        livenessProbe:
          httpGet:
            path: /livenesscheck.txt
            port: 80
          initialDelaySeconds: 15
          # Retry the probe every X seconds (frequency):
          periodSeconds: 30
          # Number of times this probe can fail,
          #   before kubernetes gives up and "restarts" the "container":
          failureThreshold: 1
      - name: liveness-with-troublemaker # Name of container in the pod.
        image: kamranazeem/k8s-probes-demo:latest

        env:
        - name: START_DELAY
          value: "120" # Double quotes for ENV variables are important.

        ports:
        - name: http 
          containerPort: 80

        resources:
          limits:
            cpu: 10m
            memory: 20Mi
          requests:
            cpu: 5m
            memory: 20Mi
       
        # Liveliness probes are used while the container is up, 
        #   to see if the service is "still" alive.
        # If the service does not respond, then kubernetes kills the container/pod,
        #   and restarts it.
        # It is therefore important to have initialDelaySeconds for the "livenessProbe"
        #   to be as late as possible.
        # If it is a small number, and if for some reason container is not started yet,
        #   (in case of slow application startup), then kubernetes will restart it.
        # This means the pod will stuck in a crash loop, and will never start.
        # Remember, livenessProbe does NOT start AFTER readiness probe.
        #   In-fact, they both start at the same time.
        # Liveness probe will vary from application to application. 
        #   Some start fast, some start slow.
        livenessProbe:
          httpGet:
            path: /livenesscheck.txt
            port: 80
          initialDelaySeconds: 15
          # Retry the probe every X seconds (frequency):
          periodSeconds: 30
          # Number of times this probe can fail,
          #   before kubernetes gives up and "restarts" the "container":
          failureThreshold: 1 
      - name: readiness-with-troublemaker # Name of container in the pod.
        image: kamranazeem/k8s-probes-demo:latest
        imagePullPolicy: Always

        env:
        - name: START_DELAY
          value: "60" # Double quotes for ENV variables are important.

        ports:
        - name: http 
          containerPort: 80

        resources:
          limits:
            cpu: 10m
            memory: 20Mi
          requests:
            cpu: 5m
            memory: 20Mi

        # Readiness probes help Kubernetes know when to start sending traffic to the pod.
        #   This probe helps kubernetes to "add" or "remove" this pod from the service in front of it.
        #   So, when the probe passes it is added to the service,
        #       when it fails at any time in it's life, kubernetes removes it from the service.
        # Restarting the "stuck" container is the job of livenessProbe, explained further down.
        # Note: Readiness probe runs throughout the life of the container - every "periodSeconds",
        #       not just at the container start time!
        #
        # The WITPASS/wordpress images takes long to start, 
        #   because themes and plugins are pulled at container start time.
        # 
        readinessProbe:
          httpGet:
            path: /readinesscheck.txt
            port: 80
          initialDelaySeconds: 5

          # Retry the probe every X seconds (frequency):
          periodSeconds: 5

          # Number of times this probe can fail ,
          #   before kubernetes gives up and marks the "pod" as "Unready":
          failureThreshold: 24

          # In the case above, the Pod will be marked "Unready",
          #   if the probe does not pass 5 x 24 = 120 seconds.
          successThreshold: 1
        
        volumeMounts:
        - name: shared-directory
          mountPath: /usr/share/nginx/html

      - name: readiness-with-troublemaker2 # Name of container in the pod.
        image: kamranazeem/k8s-probes-demo:latest
        imagePullPolicy: Always

        env:
        - name: START_DELAY
          value: "120" # Double quotes for ENV variables are important.

        ports:
        - name: http 
          containerPort: 80

        resources:
          limits:
            cpu: 10m
            memory: 20Mi
          requests:
            cpu: 5m
            memory: 20Mi

        # Readiness probes help Kubernetes know when to start sending traffic to the pod.
        #   This probe helps kubernetes to "add" or "remove" this pod from the service in front of it.
        #   So, when the probe passes it is added to the service,
        #       when it fails at any time in it's life, kubernetes removes it from the service.
        # Restarting the "stuck" container is the job of livenessProbe, explained further down.
        # Note: Readiness probe runs throughout the life of the container - every "periodSeconds",
        #       not just at the container start time!
        #
        # The WITPASS/wordpress images takes long to start, 
        #   because themes and plugins are pulled at container start time.
        # 
        readinessProbe:
          httpGet:
            path: /readinesscheck.txt
            port: 80
          initialDelaySeconds: 5

          # Retry the probe every X seconds (frequency):
          periodSeconds: 5

          # Number of times this probe can fail ,
          #   before kubernetes gives up and marks the "pod" as "Unready":
          failureThreshold: 24

          # In the case above, the Pod will be marked "Unready",
          #   if the probe does not pass 5 x 24 = 120 seconds.
          successThreshold: 1
        
        volumeMounts:
        - name: shared-directory
          mountPath: /usr/share/nginx/html
      volumes:
      - name: shared-directory
        emptyDir: {}

```

superbrothers:
Thanks for the comment! We'll be checking your PR in a few days!

heyjcollins:
bumping this PR

superbrothers:
I will work on it in the near future :bow:

superbrothers (approved):

PR #182, opened 2022-05-16
Use Go 1.18

superbrothers:
This PR changes to use Go 1.18.


superbrothers:
Thanks for the quick review! However, it seems to fail to build goreleaser. I will change to install it with `go install`.

PR #190, opened 2022-09-23
added output modes to colorize json output

prune998:
This PR is about adding two new output modes to Stern:

- `extjson` is an extended json mode where the whole output is embeded in json and the namespace/pod/container values are color-coded as in the default mode.
- `ppextjson` is the same as the above with pretty-print, so each field is on its own line.

Both of those modes can be piped into `jq` if you want full pretty-print, but you'll lose the original color-coding.

I think those options are usefull when you're dealing with Json structured logs. It helps visializing the informations without being too intrusive. 
It's better than using the `raw` output or `raw | jq '.'` as pretty-printing the whole log line can quickly fill the console if you have lots of logs.

It's not a revolution but I guess it does not hurt to add this feature :)

ex:

```shell
./stern_cli nodepool-labeler-controller-manager --tail 2  --all-namespaces --output extjson -c manager

./stern_cli nodepool-labeler-controller-manager --tail 2  --all-namespaces --output extjson -c manager | jq '.'

./stern_cli nodepool-labeler-controller-manager --tail 2  --all-namespaces --output ppextjson -c manager
```
__stern_cli_nodepool-labeler-controller-manager_--tail_2_--all-namespaces___-_and_prune_nas____and_cli_go_—_stern

__stern_cli_nodepool-labeler-controller-manager_--tail_2_--all-namespaces___-

__stern_cli_nodepool-labeler-controller-manager_--tail_2_--all-namespaces___-_and_prune_nas___

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -229,13 +229,13 @@ func (o *options) sternConfig() (*stern.Config, error) {
 		case "json":
 			t = "{{json .}}"
 		case "extjson":
-			t = "\"pod\": \"{{color .PodColor .PodName}}\", \"container\": \"{{color .ContainerColor .ContainerName}}\", \"msg\": {{extjson .Message}}"
+			t = "\"pod\": \"{{color .PodColor .PodName}}\", \"container\": \"{{color .ContainerColor .ContainerName}}\", \"message\": {{extjson .Message}}"
 			if o.allNamespaces {
 				t = fmt.Sprintf("\"namespace\": \"{{color .PodColor .Namespace}}\", %s", t)
 			}
 			t = fmt.Sprintf("{%s}", t)
 		case "ppextjson":
-			t = "  \"pod\": \"{{color .PodColor .PodName}}\",\n  \"container\": \"{{color .ContainerColor .ContainerName}}\",\n  \"msg\": {{extjson .Message}}"
+			t = "  \"pod\": \"{{color .PodColor .PodName}}\",\n  \"container\": \"{{color .ContainerColor .ContainerName}}\",\n  \"message\": {{extjson .Message}}"
 			if o.allNamespaces {
 				t = fmt.Sprintf("  \"namespace\": \"{{color .PodColor .Namespace}}\",\n%s", t)
 			}
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"regexp"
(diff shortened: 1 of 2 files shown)


rkmathi (approved):
@prune998 
Thank you for your implementation!
I think this feature is very useful.

@superbrothers @floryut 
What do you think?

cmd/cmd.go
```diff
@@ -227,6 +228,18 @@ func (o *options) sternConfig() (*stern.Config, error) {
...
 		case "json":
 			t = "{{json .}}"
+		case "extjson":
+			t = "\"pod\": \"{{color .PodColor .PodName}}\", \"container\": \"{{color .ContainerColor .ContainerName}}\", \"msg\": {{extjson .Message}}"
+			if o.allNamespaces {
+				t = fmt.Sprintf("\"namespace\": \"{{color .PodColor .Namespace}}\", %s", t)
+			}
+			t = fmt.Sprintf("{%s}", t)
+		case "ppextjson":
+			t = "  \"pod\": \"{{color .PodColor .PodName}}\",\n  \"container\": \"{{color .ContainerColor .ContainerName}}\",\n  \"msg\": {{extjson .Message}}"
```
superbrothers, line 238: I wanted to use `json.Marshal`, but found that using it would reorder the keys and make it difficult to see.

cmd/cmd.go
```diff
@@ -227,6 +228,18 @@ func (o *options) sternConfig() (*stern.Config, error) {
 			t = "{{.Message}}"
 		case "json":
 			t = "{{json .}}"
+		case "extjson":
+			t = "\"pod\": \"{{color .PodColor .PodName}}\", \"container\": \"{{color .ContainerColor .ContainerName}}\", \"msg\": {{extjson .Message}}"
```
superbrothers, line 232: I want to change `msg` to `message`. It should not be abbreviated.

```suggestion
			t = "\"pod\": \"{{color .PodColor .PodName}}\", \"container\": \"{{color .ContainerColor .ContainerName}}\", \"message\": {{extjson .Message}}"
```

floryut (approved):
> @prune998 Thank you for your implementation! I think this feature is very useful.
> 
> @superbrothers @floryut What do you think?

Great! Thank you for your contrib @prune998 
please review @superbrothers comments 👍

prune998:
All done ! Please merge if you think it's ready !

Many thanks all for supporting Stern !

superbrothers (approved):

PR #193, opened 2022-09-29
Fix the invalid command in README.md

superbrothers:
This PR fixes the invalid command in README.md. pod-query (`.` and `backend`) appears twice in that command.


rkmathi (approved):
😺

floryut (approved):
👍

superbrothers:
I do not run CI jobs when only Markdown files are changed, and it seems that the expected checks are not met in that case. This time, I will merge this PR as is.

https://github.com/stern/stern/blob/f6f4f640be851ccbd311b609d94a6fdab432d212/.github/workflows/ci.yaml#L9

Issue #15, opened 2020-10-22
Feature request: print all available logs and exit


superbrothers:
**Note:** this issue was imported from https://github.com/wercker/stern/issues/72, but it was originally created by [emptywee](https://github.com/emptywee)...

Hello.

Is it possible to add a flag to stern to stop tailing after all available logs are shown? Would be very useful when you need to print logs from multiple different pods (using a regexp or wildcard or label selector) and pipe them to grep. Currently it keeps tailing no matter what and you don't really know if stern has shown all recent logs or still printing, because grep is filtering that info out.

Thank you.

superbrothers:
**Note:** this comment was imported, but it was originally made by [hatchan](https://github.com/hatchan)...

Currently it is only possible to follow the logs, so it won't quite after it has shown the logs.

superbrothers:
**Note:** this comment was imported, but it was originally made by [olli-miettinen-arm](https://github.com/olli-miettinen-arm)...

This would be very nice feature, a stop switch that would allow using stern in scripts.

superbrothers:
**Note:** this comment was imported, but it was originally made by [henricook](https://github.com/henricook)...

We'd also love this, so we can tail multiple pods and then operate on the output (e.g. grep response times and then `| sort -n`)

superbrothers:
**Note:** this comment was imported, but it was originally made by [fuhbar](https://github.com/fuhbar)...

Same here. We'd love to see this implemented very much like `tail(1)`.

Our use case is as follows: Using helm deployment hooks in CD pipelines leaves us with the task of feeding back the pod logs to the user respectively the CD system, so that the user is able to find information why her deployment has failed.  

Stern is the perfect answer for accomplishing this. :)

superbrothers:
**Note:** this comment was imported, but it was originally made by [jacobstr](https://github.com/jacobstr)...

One nice way to accomplish this might be an `until` flag to complement `since`. e.g.

```
stern --until=$(date +%s)
```

Could print all logs until `now` and exit.

superbrothers:
**Note:** this comment was imported, but it was originally made by [ljohnston](https://github.com/ljohnston)...

IMO, a `--until` flag is the best way to implement this. Not only could you stop it at `now`, but you could go back and see what was happening during a specific time frame (e.g. `stern --since 4h --until 3h ...`).

bigbits22:
any news on this?

PR #204, opened 2023-01-13
Add --no-follow flag to exit when all logs have been shown

tksm:
Fixes https://github.com/stern/stern/issues/15

This PR adds `--no-follow` flag to exit when all logs have been shown.

The name of the flag could be `--follow=false`. But it will be used with `false` in most cases, so I chose `--no-follow` as the auto-completion can work in one shot.

### DIFF
--- a/stern/watch.go
+++ b/stern/watch.go
@@ -32,7 +32,7 @@ import (
 // Watch starts listening to Kubernetes events and emits modified
 // containers/pods. The first result is targets added, the second is targets
 // removed
-func Watch(ctx context.Context, i v1.PodInterface, labelSelector labels.Selector, fieldSelector fields.Selector, filter *targetFilter) (chan *Target, chan *Target, error) {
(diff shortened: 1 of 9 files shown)


stern/watch.go
```diff
@@ -79,79 +64,19 @@ func Watch(ctx context.Context, i v1.PodInterface, podFilter *regexp.Regexp, exc
...
-						removed <- &Target{
-							Node:      pod.Spec.NodeName,
-							Namespace: pod.Namespace,
-							Pod:       pod.Name,
-							Container: cn,
-						}
-					}
+					filter.visit(pod, func(t *Target, containerStateMatched bool) {
+						removed <- t
+					})
```
tksm, line 79: Containers without their statuses are never added, so now we get container names from container statuses in the same manner as handling Added/Modified events to simplify the logic.

superbrothers (approved):
Thank you for your contribution!😊

Issue #159, opened 2021-09-19
Add an option for a "deployment-query"


superbrothers:
**What would you like to be added**:

> The default argument takes a "pod-query", can there be an option for a "deployment-query"? 😄
> 
> That way one could pass "app$" and have it match the app-756d764c46-njkvx pod without the longer "app-[0-9a-z]{10}-[0-9a-z]{5}" regex.
> 
> Agreed, the easiest way to do this is to use a label selector. Just so I can use this tool easier, I have add something extra. Someone will ask: "Why are you adding these labels?" The answer being "Because I want to use stern in an easier way".

- https://github.com/stern/stern/issues/158#issuecomment-922359537

**Why is this needed**:

cc/ @vsimon

superbrothers:
I don't have a strong motivation to add this feature, since we can check the pod labels with `kubectl get po --show-labels`. Also, in many scenarios, pod-query gives the expected results.

```
$ kubectl create deploy app --image nginxinc/nginx-unprivileged
$ kubectl create deploy app-admin --image nginxinc/nginx-unprivileged
$ kubectl create deploy app-sdk --image nginxinc/nginx-unprivileged

$ kubectl get po --show-labels
NAME                        READY   STATUS    RESTARTS   AGE   LABELS
app-756d764c46-njkvx        1/1     Running   0          47h   app=app,pod-template-hash=756d764c46
app-admin-dc858d4d4-wmbh7   1/1     Running   0          47h   app=app-admin,pod-template-hash=dc858d4d4
app-sdk-55cbbb55-65tz4      1/1     Running   0          47h   app=app-sdk,pod-template-hash=55cbbb55

$ stern -l app=app
```

If we add a new feature, we can consider specifying the resource in the form `<resource>/<regexp>` as one idea.

```
$ stern "deployment/app$"
```

vsimon:
For my user-defined helm charts, I can add `app=<app>` (I don't think helm adds them automatically "for free" when doing helm install...). I also use third party charts, some use a different labelling convention by default like, `app.kubernetes.io/name=<app>` or `app.kubernetes.io/component=<app>`, so I'd have to add labels `app=<app>` to those kinds of charts use stern in a more uniform way.

> If we add a new feature, we can consider specifying the resource in the form <resource>/<regexp> as one idea. `$ stern "deployment/app$"`

[...]

PR #206, opened 2023-01-17
Add --resource flag to select pods by resource

tksm:
Fixes https://github.com/stern/stern/issues/159

This PR adds `--resource` flag to select pods by the specified resource in the form "\<resource\>/\<name\>".

This will be useful when we want to distinct pods created by multiple deployments whose names are partly the same.

```
$ kubectl get deploy -n kube-system | grep cert-manager
cert-manager                             1/1     1            1           61d
cert-manager-cainjector                  1/1     1            1           61d
cert-manager-webhook                     1/1     1            1           61d

# It is not easy to select only pods created by `deployment/cert-manager` with pod-query.
$ stern -n kube-system --tail 0 cert-manager
+ cert-manager-webhook-d5568c87d-tcq6g › cert-manager
+ cert-manager-84bd74df78-zb28x › cert-manager
+ cert-manager-cainjector-6dd88998fb-wpb8h › cert-manager

# The resource flag allows users to select only pods created by `deployment/cert-manager`.
$ stern -n kube-system --tail 0 --resource deploy/cert-manager
+ cert-manager-84bd74df78-zb28x › cert-manager
```

It could be the first argument in the form "\<resource\>/\<regexp\>" instead of adding the flag as in https://github.com/stern/stern/issues/159#issuecomment-922684992. But the flag with the exact match is much simpler, so I implemented it as a flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ The `pod` query is a regular expression so you could provide `"web-\w"` to tail
  `--no-follow`               | `false`   | Exit when all logs have been shown.
  `--output`, `-o`            | `default` | Specify predefined template. Currently support: [default, raw, json, extjson, ppextjson]
  `--prompt`, `-p`            | `false`   | Toggle interactive prompt for selecting 'app.kubernetes.io/instance' label values.
+ `--resource`                |           | Select pods by the specified resource in the form "&lt;resource&gt;/&lt;name&gt;". If present, default to ".*" for the pod-query.
  `--selector`, `-l`          |           | Selector (label query) to filter on. If present, default to ".*" for the pod-query.
  `--since`, `-s`             | `48h0m0s` | Return logs newer than a relative duration like 5s, 2m, or 3h.
  `--tail`                    | `-1`      | The number of lines from the end of the logs to show. Defaults to -1, showing all logs.
@@ -180,6 +181,11 @@ stern frontend --selector release=canary
 Tail the pods on `kind-control-plane` node across all namespaces
 ```
 stern --all-namespaces --field-selector spec.nodeName=kind-control-plane
+```
+
+Tail the pods created by `deployment/nginx`
+```
+stern --resource deployment/nginx
 ```
 
 Pipe the log message to jq:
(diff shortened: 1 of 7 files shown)


superbrothers:
I'll check this later😍 Thanks!

tksm:
Sorry, I found cron jobs can have no labels. I will add an error handling case when their labels are empty.

tksm:
I added an error handling the case when labels are empty in https://github.com/stern/stern/pull/206/commits/7b9aebc69c782b997f3a75855f03b9a5526399c5 .

```
$ stern --resource cronjob/hello
Error: resource cronjob/hello has no labels
```

tksm:
I will create another PR for a dynamic completion of `--resource` if this PR is approved. It will work like the demo below.

[asciicast](https://asciinema.org/a/uNoMy4K4pk9Lr6UlK1p41uvzq)

superbrothers:
This is a great addition. Thank you.

I have one point I would like to discuss. Whether or not to add `--resource` flag. Currently the argument is pod-query, but if it contains `/` it can also be considered a resource.

```console
# pod-query
stern nginx

# resource
stern deployment/nginx
```

I'm not sure if adding the `--resource` flag would be clearer. What do you think?

tksm:
@superbrothers Though it is easier to use to take a resource as the first argument indeed, I prefer it as a flag because users might expect it to take a regular expression when in the first argument.

```
stern "deployment/app.*"
```

Since it is an exact match and works as a label selector like `--selector`, I think it is natural to be a flag.

tksm:
>  I prefer it as a flag because users might expect it to take a regular expression when in the first argument.

@superbrothers On second thought, I realized the situation when users use a regular expression for a deployment is already covered by the pod query. Since it is a new feature and the current flag is mutually exclusive with the pod query, I will reconsider it as the first argument.

tksm:
@superbrothers After reimplementing it as the first argument, I feel `stern <resource>/<name>` is more comfortable than a flag. If you have no objection, I will close this PR and make a new PR to support a resource as the argument.

Thank you for your proposal. 👍

superbrothers:
I have no objection. You can re-create a PR or make changes to this PR. I'm fine with either.

tksm:
I reimplemented it as a query with some fixes in https://github.com/stern/stern/pull/208.

PR #210, opened 2023-01-22
Fix bug where container-state is ignored when no-follow specified

tksm:
This PR fixes a bug where `--container-state` is ignored when `--no-follow` is specified.


rkmathi (approved):
LGTM 😺

PR #218, opened 2023-01-28
Allow to specify --exclude-pod/container multiple times

### LINKED ISSUE #212: Allow to specify `--exclude-pod` multiple times

```
stern '.*' --exclude-pod pod-a --exclude-pod pod-b
```

So both gets excluded.

**Why is this needed**:

Currently the behavior is that only the last `--exclude-pod` specified is taken into account, the rest is ignored.

superbrothers:
This PR allows users to specify `--exclude-pod/container` multiple times.

```
$ stern . --exclude-pod pod1 --exclude-pod pod2

$ stern . --exclude-container container1 --exclude-container container2
```

Closes #212


rkmathi (approved):
LGTM 😄

Issue #217, opened 2023-01-28
Stern shows the same logs multiple times when a container is terminated


tksm:
**What happened**:

Stern shows the same logs multiple times when a container is terminated.

```
$ stern --container-state running --container-state terminated myjob
+ myjob-xz9vh › myjob
myjob-xz9vh myjob done
- myjob-xz9vh › myjob
+ myjob-xz9vh › myjob
myjob-xz9vh myjob done
- myjob-xz9vh › myjob
+ myjob-xz9vh › myjob
myjob-xz9vh myjob done
```

**What you expected to happen**:

Logs should be shown only once.

**How to reproduce it (as minimally and precisely as possible)**:

1. Run stern with `--container-state` flags.

```
stern --container-state running --container-state terminated myjob
```

2. Create a job that will immediately complete with a log "done".

```
kubectl create job myjob --image busybox -- echo "done"
```

3. The same logs are shown multiple times.

```
$ stern --container-state running --container-state terminated myjob
+ myjob-xz9vh › myjob
myjob-xz9vh myjob done
- myjob-xz9vh › myjob
+ myjob-xz9vh › myjob
myjob-xz9vh myjob done
- myjob-xz9vh › myjob
+ myjob-xz9vh › myjob
myjob-xz9vh myjob done
```

**Anything else we need to know?**:

[The tails of terminated containers will be inactive immediately](https://github.com/stern/stern/blob/v1.22.0/stern/stern.go#L128-L133), so they will be added back again when stern handles [modified events](https://github.com/stern/stern/blob/v1.22.0/stern/watch.go#L91). I think that is why the same logs are shown multiple times.

Maybe stern should ignore `terminated` containers on modified events.

**Environment**:
- stern version (use `stern --version`): 1.22.0
- OS (e.g: `cat /etc/os-release`): macOS 12.2
- Install tools (e.g: Homebrew) Homebrew
- Others:

tksm:
It might be a significant change around the pod event handling to fix the issue.

> Maybe stern should ignore terminated containers on modified events.

I found it would not work. When a pod immediately completes, it seems that the container state transition skips the "running" state, and the "waiting" state can transit to the "terminated" directly (at least from the stern's point of view).

tksm:
Similar problems also happen when a pod restarts.

The pod below will show a random number and then exit immediately.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: restart
spec:
  containers:
  - image: busybox:1.36.0
    name: busybox
[...]

PR #221, opened 2023-02-02
Improve handling of container termination

tksm:
Fixes https://github.com/stern/stern/issues/217

This PR improves the handling of container termination. As discussed in #217, stern currently does not handle container termination well. [The steps in the comment](https://github.com/stern/stern/issues/217#issuecomment-1411656570) can reproduce this issue.

This PR makes the following changes to improve.  (sorry for the big change)

- Manage the last shown Container IDs
    - Start tailing logs only when Container IDs are changed
    - It prevents the same log lines are unnecessarily shown
    - The logic to choose an ID is based on [Kubelet's validateContainerLogStatus](https://github.com/kubernetes/kubernetes/blob/v1.26.1/pkg/kubelet/kubelet_pods.go#L1246).
- The container states filter is applied only to containers that have existed before stern starts
    - It allows stern to show a pod that immediately completes that skips a "running" state.
- Do not cancel tailing when containers are terminated or removed
    - It prevents the last logs from being omitted by canceling.
    - Tailing will eventually stop by EOF or other errors
- Retry tailing if it gets an error
    - Before this change, retrying would happen when pod events fire
    - After this change, we need explicit retrying because we ignore pod events when Container IDs are not changed

This change is significant and hard to debug, so I would like to add debug logs. It also helps us to understand container state transitions. Here is an example of debug logs for [a restarted pod](https://github.com/stern/stern/issues/217#issuecomment-1411656570).

```
$ dist/stern --verbosity 7 restart 2> >(grep -E 'target.go|›')
I0202 21:21:21.051218   26997 target.go:139] "Container ID is empty" state="waiting" target="default-restart-busybox"
I0202 21:21:21.676584   26997 target.go:158] "Container ID was changed" state="terminated" target="default-restart-busybox" container="containerd://2a2a0574f0fdfbce8dddc647a1e3469e5b2ccd4e5fb8c95225e778c8f020ab04" last=""
+ restart › busybox
restart busybox 21567
- restart › busybox
I0202 21:21:22.679503   26997 target.go:158] "Container ID was changed" state="terminated" target="default-restart-busybox" container="containerd://fac6a6ad2e8029b97b9eb4a65a9168c2bea9fb063abe823749520952216a9bf0" last="containerd://2a2a0574f0fdfbce8dddc647a1e3469e5b2ccd4e5fb8c95225e778c8f020ab04"
+ restart › busybox
restart busybox 601
- restart › busybox
I0202 21:21:23.686007   26997 target.go:153] "Container ID is the same" state="waiting" target="default-restart-busybox" container="containerd://fac6a6ad2e8029b97b9eb4a65a9168c2bea9fb063abe823749520952216a9bf0"
I0202 21:21:38.713079   26997 target.go:158] "Container ID was changed" state="terminated" target="default-restart-busybox" container="containerd://d7e9e5fbc788cb67081b81697b9817ac2128a2d45e09c8e55c5095d2bf0f1128" last="containerd://fac6a6ad2e8029b97b9eb4a65a9168c2bea9fb063abe823749520952216a9bf0"
+ restart › busybox
restart busybox 19817
- restart › busybox
```

### DIFF
--- a/stern/stern.go
+++ b/stern/stern.go
@@ -19,40 +19,19 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
-	"sync"
+	"time"
 
 	"github.com/pkg/errors"
 
 	"github.com/stern/stern/kubernetes"
 	"golang.org/x/sync/errgroup"
+	"golang.org/x/time/rate"
 
 	"k8s.io/apimachinery/pkg/labels"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 )
-
-var tails = make(map[string]*Tail)
-var tailLock sync.RWMutex
-
-func getTail(targetID string) (*Tail, bool) {
-	tailLock.RLock()
-	defer tailLock.RUnlock()
-	tail, ok := tails[targetID]
-	return tail, ok
-}
-
-func setTail(targetID string, tail *Tail) {
-	tailLock.Lock()
-	defer tailLock.Unlock()
-	tails[targetID] = tail
-}
-
-func clearTail(targetID string) {
-	tailLock.Lock()
-	defer tailLock.Unlock()
-	delete(tails, targetID)
-}
 
 // Run starts the main run loop
 func Run(ctx context.Context, config *Config) error {
@@ -103,7 +82,7 @@ func Run(ctx context.Context, config *Config) error {
 		}
 	}
 
-	filter := &targetFilter{
+	filter := newTargetFilter(targetFilterConfig{
 		podFilter:              config.PodQuery,
 		excludePodFilter:       config.ExcludePodQuery,
 		containerFilter:        config.ContainerQuery,
@@ -111,7 +90,7 @@ func Run(ctx context.Context, config *Config) error {
 		initContainers:         config.InitContainers,
 		ephemeralContainers:    config.EphemeralContainers,
 		containerStates:        config.ContainerStates,
-	}
+	})
 	newTail := func(t *Target) *Tail {
 		return NewTail(client.CoreV1(), t.Node, t.Namespace, t.Pod, t.Container, config.Template, config.Out, config.ErrOut, &TailOptions{
 			Timestamps:   config.Timestamps,
@@ -155,11 +134,9 @@ func Run(ctx context.Context, config *Config) error {
 	}
 
 	added := make(chan *Target)
-	removed := make(chan *Target)
 	errCh := make(chan error)
 
 	defer close(added)
-	defer close(removed)
 	defer close(errCh)
 
 	for _, n := range namespaces {
@@ -167,7 +144,7 @@ func Run(ctx context.Context, config *Config) error {
 		if err != nil {
 			return err
 		}
-		a, r, err := WatchTargets(ctx,
+		a, err := WatchTargets(ctx,
 			client.CoreV1().Pods(n),
 			selector,
 			config.FieldSelector,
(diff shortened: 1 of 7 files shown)


tksm:
📝 This change will also allow users to use stern to debug CrashLoopBackoff containers.

superbrothers:
Thanks. I will check this out this weekend!

stern/stern.go
```diff
@@ -185,50 +161,41 @@ func Run(ctx context.Context, config *Config) error {
...
+				return
+			}
+			if interval := time.Until(start.Add(1 * time.Minute)); interval > 0 {
+				fmt.Fprintf(config.ErrOut, "failed to tail: %v, will retry in %s\n", err, interval)
+				select {
+				case <-ctx.Done():
+				case <-time.After(interval):
 				}
-			}(tail)
+				return
```
superbrothers, line 190: Since `addTarget` function exits here, it does not look like it is retrying.

stern/stern.go
```diff
@@ -185,50 +161,41 @@ func Run(ctx context.Context, config *Config) error {
...
-			setTail(targetID, tail)
-
-			go func(tail *Tail) {
-				if err := tail.Start(ctx); err != nil {
-					fmt.Fprintf(config.ErrOut, "unexpected error: %v\n", err)
+			if !filter.isActive(target) {
+				fmt.Fprintf(config.ErrOut, "failed to tail: %v\n", err)
+				return
+			}
+			if interval := time.Until(start.Add(1 * time.Minute)); interval > 0 {
```
superbrothers, line 184: What should we do with the retry interval? I don't think the error occurs frequently, so I think it's a problem that doesn't need to be considered deeply. Can we consider an exponential backoff of up to 1 minute?

stern/stern.go
tksm: That's true. I will fix it after https://github.com/stern/stern/pull/221#discussion_r1096703044. Thanks!

stern/stern.go
tksm: Thank you for pointing it out.

There would be two cases when we need to retry.

1. It fails when it starts
    - e.g., lack of the `pods/log` permission
2. It is disconnected on the way
    - Load balancers or firewalls might disconnect when there are no logs for a while

We need to prevent a burst of retries in case 1. On the other hand, I think it is better to retry immediately in case 2. (Case 2 often happens in my use case :cry:.)

What do you think about using [a rate limiter](https://pkg.go.dev/golang.org/x/time/rate) to support the above two cases? We can consider that if it takes 10 seconds or more, it is case 2.

https://go.dev/play/p/tdNp9huQmSI

```go
	addTarget := func(ctx context.Context, target *Target) {
		limiter := rate.NewLimiter(rate.Every(time.Second*10), 3)
		for {
			if err := limiter.Wait(ctx); err != nil {
				fmt.Fprintf(config.ErrOut, "failed to retry: %v\n", err)
				return
			}
			tail := newTail(target)
			err := tail.Start(ctx)
			tail.Close()
			if err == nil {
				return
			}
			if !filter.isActive(target) {
				fmt.Fprintf(config.ErrOut, "failed to tail: %v\n", err)
				return
			}
			fmt.Fprintf(config.ErrOut, "failed to tail: %v, will retry\n", err)
		}
	}
```

superbrothers: rate limiter looks good!

tksm: Thank you! I changed it in https://github.com/stern/stern/pull/221/commits/01dedcad56fbb135043e7ae8c71fba98ca38d1d8 .

stern/stern.go
tksm: I've fixed it in https://github.com/stern/stern/commit/01dedcad56fbb135043e7ae8c71fba98ca38d1d8.

tksm:
I'll rebase and squash after the review.

superbrothers (approved):

tksm:
Thank you for the review. I've rebased on master in https://github.com/stern/stern/commit/a98c0b4eded10629b6308e8de71c26ca10b3e983.

PR #223, opened 2023-02-04
Allow pods without labels to be selected in the resource query

tksm:
This PR allows pods without labels to be selected in the resource query.

Currently, we will get the following error when we specify a pod without labels.

```
$ stern pod/nolabel
Error: resource pod/nolabel has no labels to select
```

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nolabel
spec:
  containers:
  - name: nginx
    image: nginx:1.23.3
```


superbrothers (approved):
Good catch!

tksm:
Thank you for the review! I rebased it in https://github.com/stern/stern/commit/518996274f6824b7dfd8babec53d6969a7126871.

Issue #73, opened 2020-10-24
Remove support for --all-namespaces and add support for multiple --namespace


ssoriche:
I work with several large clusters with a lot of namespaces. Running stern and specifying `--all-namespaces` is potentially costly and dangerous to the health of the cluster. I would like to remove the option to specify `--all-namespaces`, and replace it with the ability to specify multiple `--namespace` options. The thinking is that there is more thought required to adding each namespace than just asking for every one.

floryut:
Well I agree that specifying namespace could be an added option but I'm strongly against removing all namespaces as it's pretty much my daily use case :)

superbrothers:
I've probably never used `--all-namespaces` before. I don't know the case where I want to log across all namespaces.

On the other hand, in an environment with thousands or tens of thousands of pods, it's true that execution with the `--all-namespaces` flag is dangerous.

floryut:
I have some use cases with resources handle by operator on our end, which create some pods in each namespace declared by users.
And sometime I want to check all those pods, I only know the name not the namespace (well I can get it of course with a `k get pod -A`) and it's very handy to be able to follow all those pods without doing anything.

Also I'm lazy and sometime I want to check api pods without having to specify `n kube-system` and only doing `-A`  :P

I'm really strongly against removing it (or even drepracating it).
But I agree that we could support multiple namespace and maybe put a warning on `all-namespace` so people understand the "risk"

superbrothers:
> And sometime I want to check all those pods, I only know the name not the namespace (well I can get it of course with a k get pod -A) and it's very handy to be able to follow all those pods without doing anything.

I understand your use case. For now, leave it as it is until more similar opinions come. :+1:

ssoriche:
I understand the use case on smaller clusters, I have one of those myself that it is just easier to use `-A` on to see what's going on, but the environment I really care about and have other users that could really benefit using stern on has a thousand namespaces and tens of thousands of pods. It wouldn't take much for an accidental `--all-namespaces` in there and do some damage.

A warning message only helps if the user can stop the process before it communicates with the API. What about a configuration option that enables the use of `--all-namespaces`?

superbrothers:
@ssoriche @floryut How about setting an upper limit on the number of pods or containers to handle? When the limit is reached, stern is terminated. This also makes the `--all-namespaces` flag safe.

What do you think about setting the maximum number of pods to 100?

floryut:
[...]

PR #224, opened 2023-02-07
Add --max-log-requests flag to limit concurrent requests

tksm:
Fixes https://github.com/stern/stern/issues/73

This PR adds the `--max-log-requests` flag to limit concurrent requests to prevent unintentional load to a cluster.

The behavior and the default are different depending on the presence of the `--no-follow` flag.

| `--no-follow` | default | behavior         |
|---------------|---------|------------------|
| specified     | 5       | limits the number of concurrent logs to request |
| not specified | 50      | exits with an error when if it reaches the concurrent limit |

## --no-follow specified
It limits the number of concurrent logs to request, so there is no apparent change.
I chose 5 for the default limit, the same as in kubectl.

:memo: I like the combination of `--max-log-requests 1` and `--no-follow`, which will show logs in order.

## --no-follow is not specified
It exits with an error when if it reaches the concurrent limit.
Though I am not sure how we should choose the default limit, I think 50 is enough for most cases.

```
$ dist/stern --max-log-requests 3 deployment/nginx
+ nginx-76d6c9b8c-77s2r › nginx
+ nginx-76d6c9b8c-cg85s › nginx
+ nginx-76d6c9b8c-lxpmt › nginx
Error: stern reached the maximum number of log requests (3), use --max-log-requests to increase the limit
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ Supported Kubernetes resources are `pod`, `replicationcontroller`, `service`, `d
  `--include`, `-i`           | `[]`      | Log lines to include. (regular expression)
  `--init-containers`         | `true`    | Include or exclude init containers.
  `--kubeconfig`              |           | Path to kubeconfig file to use. Default to KUBECONFIG variable then ~/.kube/config path.
+ `--max-log-requests`        | `-1`      | Maximum number of concurrent logs to request. Defaults to 50, but 5 when specifying --no-follow
  `--namespace`, `-n`         |           | Kubernetes namespace to use. Default to namespace configured in kubernetes context. To specify multiple namespaces, repeat this or set comma-separated value.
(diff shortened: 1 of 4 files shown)


rkmathi (approved):
LGTM 😸

felipecrs:
How to remove the limit, -1?

tksm:
@felipecrs Currently, stern does not support removing the requests limit. You need to specify a big enough number of the limit, like `--max-log-requests 100000`, if you want to avoid the limit.

It is a bit bothersome. https://github.com/stern/stern/issues/252 might help to configure the default of the request limit in the future.

superbrothers:
@felipecrs We've released a new version that includes support for config file. Please try it out.

```
$ mkdir -p ~/.config/stern
$ cat <<EOL
max-log-requests: 100000
EOL
$ stern .
```

Issue #36, opened 2020-10-22
Show all logs from all containers from all states together including previous containers


superbrothers:
**Note:** this issue was imported from https://github.com/wercker/stern/issues/117, but it was originally created by [kivagant-ba](https://github.com/kivagant-ba)...

Related to:
- https://github.com/wercker/stern/issues/72
- https://github.com/wercker/stern/issues/103
- https://github.com/wercker/stern/issues/68

Stern is very useful and great but there are some cases where it requires additional commands.
It would be nice to have the only one option to print-and-exit _everything_ that matches the given filter:

- logs from active pods and all their containers inside.
- logs from terminated/exited pods (completed jobs) and all their containers.
- logs from all previous containers that crashed/exited and restarted by Kubernetes.
- everything else that Stern can support.

So the feature request combines other issues together to get all available logs without using any parameters combinations.

Use case:
1. Install a Helm chart without analyzing its resources.
2. With `stern` print logs from everything installed based on the project name including jobs, crashed pods, running pods etc.

superbrothers:
**Note:** this comment was imported, but it was originally made by [rmetzler](https://github.com/rmetzler)...

I would love to see logs from completed pods. That way I could stern to view logs from cronjobs.

superbrothers:
**Note:** this comment was imported, but it was originally made by [hulucc](https://github.com/hulucc)...

And it's useful when debugging problems

superbrothers:
**Note:** this comment was imported, but it was originally made by [brettplarson](https://github.com/brettplarson)...

This would be great!

DTTerastar:
Not showing terminated pods JUST bit me when demoing stern.  It should not default that way.  The most common case for looking for logs to figure out why it started and immediately died!

igortavtib:
I think this issue was solved by https://github.com/stern/stern/pull/222

PR #225, opened 2023-02-08
Change the default of --container-state to `all`

tksm:
This PR changes the default value of --container-state to [`all`](https://github.com/stern/stern/pull/222). It will be a breaking change.

Currently, the default value is `running`, so it does not show logs of completed pods (`terminated`) and CrashLoopBackoff pods (`waiting`). It is a bit [confusing](https://github.com/stern/stern/issues/36#issuecomment-1108906137) because `kubectl logs` shows logs regardless of their container statuses.

I don't know the exact reason why the default is `running`, but I guess it is because stern did not handle terminated containers well. By #221, stern can now handle terminated containers, so I would like to consider changing the default to `all`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ Supported Kubernetes resources are `pod`, `replicationcontroller`, `service`, `d
  `--color`                   | `auto`    | Force set color output. 'auto':  colorize if tty attached, 'always': always colorize, 'never': never colorize.
  `--completion`              |           | Output stern command-line completion code for the specified shell. Can be 'bash', 'zsh' or 'fish'.
  `--container`, `-c`         | `.*`      | Container name when multiple containers in pod. (regular expression)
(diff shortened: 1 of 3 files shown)


rkmathi (commented):
This is indeed a breaking change, but I think it useful.

@superbrothers @floryut 
What do you think?

superbrothers (approved):
I am concerned that if there are many Pods created from Jobs, their log output will make it difficult to see the desired log. However, this should not be a problem since we can simply change pod-query or container-state.

tksm:
@floryut Do you have any concerns about this change? It will be a breaking change, so I would like to get a consensus from all maintainers.

tksm:
@floryut I'll merge this PR once but feel free to comment if you have any concerns about this change.

floryut:
Sorry I was away for 10 days 🎿 

I'm fine with this, this is a breaking change, but I'll think it's better to follow the same logic as kubectl as long as the logic is not buggy/faulty

Issue #227, opened 2023-02-11
Add support to parse custom JSON logs


niamster:
**What would you like to be added**:

I'd like to add a support for parsing logs in JSON format and re-format for better human perception.
Particularly I'm interested in logs generated by https://github.com/uber-go/zap

**Why is this needed**:
Everyone is leaning to generate logs in JSON, this is great, it's easier to parse them for machines. 
This is less human-friendly though.
And `stern` is for humans, even if application logs are machine-friendly.

PR #228, opened 2023-02-11
add support to parse JSON logs

niamster:
Addressed #227

This PR adds:
* add --input-format option that allows specifying message input format
* add `zap-json` input format support


cmd/cmd.go
```diff
@@ -234,24 +236,36 @@ func (o *options) sternConfig() (*stern.Config, error) {
 		return nil, errors.New("color should be one of 'always', 'never', or 'auto'")
 	}
 
+	messageTpl := "{{.Message}}"
+	switch o.inputFormat {
+	case "zap-json":
+		if color.NoColor {
+			messageTpl = `{{ with $msg := .Message | parseJSON }}[{{ $msg.ts }}] {{ $msg.level }} ({{ $msg.caller }}) {{ $msg.msg }}{{ end }}`
+		} else {
+			messageTpl = `{{ with $msg := .Message | parseJSON }}[{{ color $.TimestampColor $msg.ts }}] {{ levelColor $msg.level }} ({{ color $.CallerColor $msg.caller }}) {{ $msg.msg }}{{ end }}`
```
superbrothers, line 245: I would like to create a generic color function and use it as follows. This will eliminate the need to change `tail.go`.

```diff
diff --git a/cmd/cmd.go b/cmd/cmd.go
index 5dc6784..4395441 100644
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -242,7 +242,7 @@ func (o *options) sternConfig() (*stern.Config, error) {
                if color.NoColor {
                        messageTpl = `{{ with $msg := .Message | parseJSON }}[{{ $msg.ts }}] {{ $msg.level }} ({{ $msg.caller }}) {{ $msg.msg }}{{ end }}`
                } else {
-                       messageTpl = `{{ with $msg := .Message | parseJSON }}[{{ color $.TimestampColor $msg.ts }}] {{ levelColor $msg.level }} ({{ color $.CallerColor $msg.caller }}) {{ $msg.msg }}{{ end }}`
+                       messageTpl = `{{ with $msg := .Message | parseJSON }}[{{ colorGreen $msg.ts }}] {{ levelColor $msg.level }} ({{ colorCyan $msg.caller }}) {{ $msg.msg }}{{ end }}`
                }
        case "default":
        default:
@@ -312,6 +312,14 @@ func (o *options) sternConfig() (*stern.Config, error) {
                "color": func(color color.Color, text string) string {
                        return color.SprintFunc()(text)
                },
+               "colorBlack":   color.BlackString,
+               "colorRed":     color.RedString,
+               "colorGreen":   color.GreenString,
+               "colorYellow":  color.YellowString,
+               "colorBlue":    color.BlueString,
+               "colorMagenta": color.MagentaString,
+               "colorCyan":    color.CyanString,
+               "colorWhite":   color.WhiteString,
                "levelColor": func(level string) string {
                        var levelColor *color.Color
                        switch strings.ToLower(level) {
```

README.md
```diff
@@ -84,6 +84,7 @@ Supported Kubernetes resources are `pod`, `replicationcontroller`, `service`, `d
  `--field-selector`          |           | Selector (field query) to filter on. If present, default to ".*" for the pod-query.
  `--include`, `-i`           | `[]`      | Log lines to include. (regular expression)
  `--init-containers`         | `true`    | Include or exclude init containers.
+ `--input-format`, `-I`      | `default` | Specify input format. Currently support: [default, zap-json]
```
superbrothers, line 87: Since the name of the option that specifies the output format is `output`, it seems that the name of the option that specifies the input format should be `input`, but I am torn.

cmd/cmd.go
niamster: Thank you for the suggestion! I was not happy about adding specific color fields into the `tail.go` file, and I was thinking more about function like `colorFromString`, or even change default `color` function to test agains specific constants, but I was not sure whether it worth the trouble.
I like your idea, will apply your suggestion 👍

README.md
niamster: I'm a little bit hesitant about `--input` since it's too broad. If one day someone would like to add `--input` option for whatever reason, that person would have hard time finding an appropriate name.

tksm:
Thank you for the PR!

I understand the use case, but I would prefer not to support application or library-specific formats to keep Stern simple. There are countless log libraries and log formats, so it will soon be unmanageable in this approach.

I believe doing the log formatting and the syntax highlighting in another program is better. It is more flexible and works better with other programs.

```
stern <QUERY> | read-zap-json
```

superbrothers:
> There are countless log libraries and log formats, so it will soon be unmanageable in this approach.

Yes, you may be right. We need to continue to manage once we support. However, since there is no change to the default, it seems to me that if it breaks, the person who needs it can just fix it.

tksm:
> However, since there is no change to the default, it seems to me that if it breaks, the person who needs it can just fix it.

That makes sense. In that case, I have no objection to adding this feature.

README.md
tksm: I feel that the flag name `--input-format` is a bit confusing because its purpose is to change the output format. We can consider it a formatter for `json-zap`, so how about `--formatter`?

```
stern --formatter zap-json <QUERY>
```

niamster: Yeah, I was afraid that `--input-format` might sound misleading, one of the option I had in mind is `--reformat` 😆 .

`--formatter` sounds good for me, @superbrothers what do you think?

niamster:
> Thank you for the PR!
> 
> I understand the use case, but I would prefer not to support application or library-specific formats to keep Stern simple. There are countless log libraries and log formats, so it will soon be unmanageable in this approach.
> 
> I believe doing the log formatting and the syntax highlighting in another program is better. It is more flexible and works better with other programs.
> 
> ```
> stern <QUERY> | read-zap-json
> ```

Thank you for taking time to review 🙇 

I perfectly understand your reasoning, and TBH I was considering using something like https://github.com/garabik/grc. At the same time it looks an overkill, and `stern` already's doing 99% of useful work.

I was thinking that at some point these templates could be moved to configuration files, and even have a dedicated "community" repo if their number starts to grow, but it's too early to make such decision.

tksm:
> I was thinking that at some point these templates could be moved to configuration files, and even have a dedicated "community" repo if their number starts to grow, but it's too early to make such decision.

I agree it's too early to move to configuration files, so I favor implementing it as a flag. :+1:

README.md
niamster: I've changed `--input-format` to `--formatter`.

niamster:
@tksm @superbrothers Thank you for taking your time for the review! Is there anything else I could do to get this PR approved and merged?

tksm (approved):
LGTM 🎉

tksm:
@superbrothers Please check the change of the flag name to `--formatter` to be sure. I'll leave it to you to merge.

superbrothers (approved):

niamster:
Thank you folks 🙇

tksm:
@niamster I tried this feature and found the following problems. Could you check them?

I assume the format of `zap-json` is as follows. https://go.dev/play/p/6UfoGdeqJ9f

```json
{"level":"info","ts":1257894000,"caller":"sandbox2210433069/prog.go:10","msg":"hello","key1":"val1"}
{"level":"info","ts":1257894000,"caller":"sandbox2210433069/prog.go:13","msg":"world","key1":"val1"}
```

When we specify `--color always`, it fails with the following errors. It might be necessary to convert `ts` to a string (or a human-readable date).

```
$ dist/stern --color always --formatter zap-json pods/zap-log --no-follow --only-log-lines
expanding template failed: template: log:1:128: executing "log" at <$msg.ts>: wrong type for value; expected string; got float64
expanding template failed: template: log:1:128: executing "log" at <$msg.ts>: wrong type for value; expected string; got float64
```

When we specify `--color never`, it works, but fields other than msg are missing. Is this what you intended?

```
$ dist/stern --color never --formatter zap-json pods/zap-log --no-follow --only-log-lines
zap-log c [1.257894e+09] info (sandbox2210433069/prog.go:10) hello
zap-log c [1.257894e+09] info (sandbox2210433069/prog.go:13) world
```

I tested with the following manifest.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: zap-log
spec:
  restartPolicy: Never
  containers:
  - command:
    - sh
    - -c
    - |
      cat <<EOS
      {"level":"info","ts":1257894000,"caller":"sandbox2210433069/prog.go:10","msg":"hello","key1":"val1"}
      {"level":"info","ts":1257894000,"caller":"sandbox2210433069/prog.go:13","msg":"world","key1":"val1"}
      EOS
    image: busybox:1.36.0
    name: c
```

niamster:
> @niamster I tried this feature and found the following problems. Could you check them?
> 
> I assume the format of `zap-json` is as follows. https://go.dev/play/p/6UfoGdeqJ9f
> 
> ```json
> {"level":"info","ts":1257894000,"caller":"sandbox2210433069/prog.go:10","msg":"hello","key1":"val1"}
> {"level":"info","ts":1257894000,"caller":"sandbox2210433069/prog.go:13","msg":"world","key1":"val1"}
> ```
> 
> When we specify `--color always`, it fails with the following errors. It might be necessary to convert `ts` to a string (or a human-readable date).
> 
> ```
> $ dist/stern --color always --formatter zap-json pods/zap-log --no-follow --only-log-lines
> expanding template failed: template: log:1:128: executing "log" at <$msg.ts>: wrong type for value; expected string; got float64
> expanding template failed: template: log:1:128: executing "log" at <$msg.ts>: wrong type for value; expected string; got float64
> ```
> 
> When we specify `--color never`, it works, but fields other than msg are missing. Is this what you intended?
> 
> ```
> $ dist/stern --color never --formatter zap-json pods/zap-log --no-follow --only-log-lines
> zap-log c [1.257894e+09] info (sandbox2210433069/prog.go:10) hello
> zap-log c [1.257894e+09] info (sandbox2210433069/prog.go:13) world
> ```
> 
> I tested with the following manifest.
> 
> ```yaml
> apiVersion: v1
> kind: Pod
> metadata:
>   name: zap-log
> spec:
>   restartPolicy: Never
>   containers:
>   - command:
>     - sh
>     - -c
>     - |
>       cat <<EOS
>       {"level":"info","ts":1257894000,"caller":"sandbox2210433069/prog.go:10","msg":"hello","key1":"val1"}
>       {"level":"info","ts":1257894000,"caller":"sandbox2210433069/prog.go:13","msg":"world","key1":"val1"}
>       EOS
>     image: busybox:1.36.0
>     name: c
> ```

Hi Takashi, I have a different `zap` output format:
```
{"level":"INFO","ts":"2023-02-20T17:56:27.238272+00:00", ...
```

You can see that the `ts` field is string and not an integer. I dug a little bit and it seems that in my environment a custom TS encoding is used for historical reasons.
Do you want me to revert this PR and propose some different approach or some fix-forward PR is better?

Thank you for your understanding!

tksm:
Thank you for confirming!

> Do you want me to revert this PR and propose some different approach or some fix-forward PR is better?

If you plan to fix it soon, either is fine for me. I prefer to revert if it takes some time because it might be included in the next release.

niamster:
> Thank you for confirming!
> 
> > Do you want me to revert this PR and propose some different approach or some fix-forward PR is better?
> 
> If you plan to fix it soon, either is fine for me. I prefer to revert if it takes some time because it might be included in the next release.

Let's revert then to avoid any problems https://github.com/stern/stern/pull/232

I will propose a new PR later when I have more time to investigate possible solutions.
Thank you for your understanding.

Issue #195, opened 2022-10-09
color the outputs with `-i`


jinleileiking:
My log is:
`bla bla bla MYTEXT bla bla`

cmd : 
` stern   -i MYTEXT`

We can filter the MYTEXT log, but `MYTEXT` text is not colored

I want to get the     

`$$\textcolor{black}{\text{bla bla bla }}\textcolor{red}{\text{MYTEXT }}\textcolor{black}{\text{ bla bla bla}}$$`

suck red text in github!

PR #231, opened 2023-02-18
Highlight matched strings in the log lines with the include option

superbrothers:
This PR changes to highlight matched strings in the log lines with the include option.

image

Closes https://github.com/stern/stern/issues/195

### DIFF
--- a/stern/tail.go
+++ b/stern/tail.go
@@ -23,6 +23,7 @@ import (
 	"hash/fnv"
 	"io"
 	"regexp"
+	"sort"
 	"strings"
 	"text/template"
 	"time"
@@ -68,6 +69,9 @@ type TailOptions struct {
 	TailLines    *int64
 	Follow       bool
 	OnlyLogLines bool
+
+	// regexp for highlighting the matched string
+	reHightlight *regexp.Regexp
 }
 
 type ResumeRequest struct {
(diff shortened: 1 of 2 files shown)


stern/tail.go
```diff
@@ -99,6 +99,25 @@ func (o TailOptions) IsInclude(msg string) bool {
...
+
+	for _, rin := range o.Include {
+		msg = rin.ReplaceAllStringFunc(msg, func(part string) string {
+			matched = true
+			return colorHighlight(part)
+		})
+
+		if matched {
+			break
+		}
```
tksm, line 115: We can specify the include option multiple times, but it seems that this logic highlights first matching regular expression only. Was this done intentionally?

image

superbrothers: Oh, I thought I should only highlight the first match, but I was wrong. I will fix it.

superbrothers: I've fixed it in 06904a6.

image

stern/tail.go
```diff
@@ -99,6 +103,30 @@ func (o TailOptions) IsInclude(msg string) bool {
 	return false
 }
 
+var colorHighlight = color.New(color.FgRed, color.Bold).SprintFunc()
+
+func (o TailOptions) HighlightMatchedString(msg string) string {
+	if o.reHightlight == nil {
```
tksm, line 109:
```suggestion
	if len(o.Include) == 0 {
		return msg
	}
	
	if o.reHightlight == nil {
```

If the include option is not specified, the following process is unnecessary, so I prefer to add an early return.

superbrothers:
I've squashed this PR.

tksm (approved):
LGTM 🎉

tksm:
📝 I like that `-i ^` allows us to highlight lines without filtering.

image

Issue #234, opened 2023-02-23
Empty lines are printed if an invalid output option is specified


tksm:
**What happened**:

Empty lines are printed if an invalid output option is specified.

```
$ stern -o invalid deployment/nginx --tail 5 --no-follow
+ nginx-748c667d99-wpxqh › nginx





- nginx-748c667d99-wpxqh › nginx
```

**What you expected to happen**:

It should return an error if an invalid output option is specified.

**How to reproduce it (as minimally and precisely as possible)**:

```
$ stern -o invalid .
```

**Anything else we need to know?**:

I'll fix it after https://github.com/stern/stern/pull/233 is merged.

**Environment**:
- stern version (use `stern --version`): 1.23.0
- OS (e.g: `cat /etc/os-release`): macOS 12.2
- Install tools (e.g: Homebrew): Homebrew
- Others:

PR #235, opened 2023-02-25
Return error when output option is invalid

tksm:
Fixes https://github.com/stern/stern/issues/234

This PR fixes a bug that prints empty lines when an invalid output option is specified.


superbrothers (approved):
:+1:

floryut:
Wow.. thank you for this, can't believe that this was missed in the first place😅

PR #237, opened 2023-02-26
Make CI jobs faster

superbrothers:
This PR makes CI jobs faster.

We have a slow CI jobs problem because they target all GOOS and GOARCH to create artifacts. Here we use `goreleaser-filter` tool to make CI jobs faster by focusing on a single target.

- https://github.com/t0yv0/goreleaser-filter


tksm (approved):
LGTM 🚀

PR #239, opened 2023-02-26
allow flexible log parsing and formatting

niamster:
Addresses https://github.com/stern/stern/issues/227

This is an alternative solution to https://github.com/stern/stern/pull/228

Instead of hardcoding template in the code, I suggest adding a bunch of helper functions and a way to load template from disk.

This PR adds:
 * various color functions
 * TS format function `formatTsRFC3339Nano`
 * `tryParseJSON` function that does not return an error
 * `--template-file/-T` to load template from disk

Sample template that I use:
```
{{color .PodColor .PodName}} {{color .ContainerColor .ContainerName}} {{ with $msg := .Message | tryParseJSON }}[{{ colorGreen (formatTsRFC3339Nano $msg.ts) }}] {{ levelColor $msg.level }} ({{ colorCyan $msg.caller }}) {{ $msg.msg }}{{ else }} {{ .Message }} {{ end }}
```

It allows me to have an alias `alias stern="stern -T ~/.stern.tpl"` which support parsing both JSON and non-JSON logs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,7 @@ stern --template='{{.PodName}}/{{.ContainerName}} {{with $d := .Message | parseJ
 Output using a custom template that tries to parse JSON or fallbacks to raw format:
 
 ```
-stern --template='{{.PodName}}/{{.ContainerName}} {{ with $msg := .Message | tryParseJSON }}[{{ colorGreen (toRFC3339Nano $msg.ts) }}] {{ levelColor $msg.level }} ({{ colorCyan $msg.caller }}) {{ $msg.msg }}{{ else }} {{ .Message }} {{ end }}{{"\n"}}' backend
+stern --template='{{.PodName}}/{{.ContainerName}} {{ with $msg := .Message | tryParseJSON }}[{{ colorGreen (toRFC3339Nano (toUTC $msg.ts)) }}] {{ levelColor $msg.level }} ({{ colorCyan $msg.caller }}) {{ $msg.msg }}{{ else }} {{ .Message }} {{ end }}{{"\n"}}' backend
 ```
 
 Load custom template from file:
--- a/README.md
+++ b/README.md
@@ -248,6 +248,18 @@ Output using a custom template with `parseJSON`:
 stern --template='{{.PodName}}/{{.ContainerName}} {{with $d := .Message | parseJSON}}[{{$d.level}}] {{$d.message}}{{end}}{{"\n"}}' backend
 ```
 
+Output using a custom template that tries to parse JSON or fallbacks to raw format:
+
+```
+stern --template='{{.PodName}}/{{.ContainerName}} {{ with $msg := .Message | tryParseJSON }}[{{ colorGreen (toRFC3339Nano $msg.ts) }}] {{ levelColor $msg.level }} ({{ colorCyan $msg.caller }}) {{ $msg.msg }}{{ else }} {{ .Message }} {{ end }}{{"\n"}}' backend
+```
+
+Load custom template from file:
+
+```
+stern --template-file=~/.stern.tpl backend
+```
+
 Trigger the interactive prompt to select an 'app.kubernetes.io/instance' label value:
 
 ```
--- a/README.md
+++ b/README.md
@@ -251,7 +251,7 @@ stern --template='{{.PodName}}/{{.ContainerName}} {{with $d := .Message | parseJ
 Output using a custom template that tries to parse JSON or fallbacks to raw format:
 
 ```
-stern --template='{{.PodName}}/{{.ContainerName}} {{ with $msg := .Message | tryParseJSON }}[{{ colorGreen (toRFC3339Nano (toUTC $msg.ts)) }}] {{ levelColor $msg.level }} ({{ colorCyan $msg.caller }}) {{ $msg.msg }}{{ else }} {{ .Message }} {{ end }}{{"\n"}}' backend
+stern --template='{{.PodName}}/{{.ContainerName}} {{ with $msg := .Message | tryParseJSON }}[{{ colorGreen (toRFC3339Nano $msg.ts) }}] {{ levelColor $msg.level }} ({{ colorCyan $msg.caller }}) {{ $msg.msg }}{{ else }} {{ .Message }} {{ end }}{{"\n"}}' backend
 ```
 
 Load custom template from file:
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ Supported Kubernetes resources are `pod`, `replicationcontroller`, `service`, `d
  `--since`, `-s`             | `48h0m0s` | Return logs newer than a relative duration like 5s, 2m, or 3h.
  `--tail`                    | `-1`      | The number of lines from the end of the logs to show. Defaults to -1, showing all logs.
  `--template`                |           | Template to use for log lines, leave empty to use --output flag.
- `--template-file`, `-T`     |           | Path to template to use for log lines, leave empty to use --output flag.
+ `--template-file`, `-T`     |           | Path to template to use for log lines, leave empty to use --output flag. It overrides --template option.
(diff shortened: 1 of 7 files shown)


cmd/cmd.go
```diff
@@ -302,6 +306,7 @@ func (o *options) AddFlags(fs *pflag.FlagSet) {
 	fs.DurationVarP(&o.since, "since", "s", o.since, "Return logs newer than a relative duration like 5s, 2m, or 3h.")
 	fs.Int64Var(&o.tail, "tail", o.tail, "The number of lines from the end of the logs to show. Defaults to -1, showing all logs.")
 	fs.StringVar(&o.template, "template", o.template, "Template to use for log lines, leave empty to use --output flag.")
+	fs.StringVarP(&o.templateFile, "template-file", "T", o.templateFile, "Path to template to use for log lines, leave empty to use --output flag.")
```
superbrothers, line 309: Please modify so that `--template` and `--template-file` options cannot be used at the same time.

README.md
```diff
@@ -95,6 +95,7 @@ Supported Kubernetes resources are `pod`, `replicationcontroller`, `service`, `d
  `--since`, `-s`             | `48h0m0s` | Return logs newer than a relative duration like 5s, 2m, or 3h.
  `--tail`                    | `-1`      | The number of lines from the end of the logs to show. Defaults to -1, showing all logs.
  `--template`                |           | Template to use for log lines, leave empty to use --output flag.
+ `--template-file`, `-T`     |           | Path to template to use for log lines, leave empty to use --output flag.
```
superbrothers, line 98: Please add examples using this change to https://github.com/stern/stern#examples.

superbrothers (commented):
Please add enough tests as needed for changes to `generateTemplat` function.

cmd/cmd.go
```diff
@@ -348,6 +360,13 @@ func (o *options) generateTemplate() (*template.Template, error) {
 			}
 			return string(b), nil
 		},
+		"tryParseJSON": func(text string) map[string]interface{} {
```
tksm, line 363: Could you add the new functions to the "templates" section in README.md?
https://github.com/stern/stern#templates

cmd/cmd.go
```diff
@@ -348,6 +360,13 @@ func (o *options) generateTemplate() (*template.Template, error) {
...
+			obj := make(map[string]interface{})
+			if err := json.Unmarshal([]byte(text), &obj); err != nil {
+				return nil
+			}
+			return obj
+		},
```
tksm, line 369:
```suggestion
		"tryParseJSON": func(text string) map[string]interface{} {
			decoder := json.NewDecoder(strings.NewReader(text))
			decoder.UseNumber()
			obj := make(map[string]interface{})
			if err := decoder.Decode(&obj); err != nil {
				return nil
			}
			return obj
		},
```

Though parseJSON has the same issue, it seems that tryParseJSON does not parse large numbers such as timestamps as expected.

Let's say we have logs below. The timestamp `1257894000` is `2009-11-10 23:00:00` in UTC.

```
{"level":"info","ts":1257894000,"caller":"sandbox2210433069/prog.go:10","msg":"hello","key1":"val1"}
{"level":"info","ts":1257894000,"caller":"sandbox2210433069/prog.go:13","msg":"world","key1":"val1"}
```

When we use `tryParseJSON` and `formatTsRFC3339Nano`, we will get `0001-01-01T00:00:00Z`, which is incorrect.

```
zap-log c [0001-01-01T00:00:00Z] info (sandbox2210433069/prog.go:10) hello
zap-log c [0001-01-01T00:00:00Z] info (sandbox2210433069/prog.go:13) world
```

Please refer to https://stackoverflow.com/questions/22343083/json-unmarshaling-with-long-numbers-gives-floating-point-number for details.

cmd/cmd.go
```diff
@@ -365,9 +384,41 @@ func (o *options) generateTemplate() (*template.Template, error) {
 			}
 			return strings.TrimSuffix(string(b), "\n"), nil
 		},
+		"formatTsRFC3339Nano": func(ts any) string {
```
tksm, line 387: I think the `Ts` part does not give additional information for users, so what do you think about simply naming it `rfc3339Nano` or `toRFC3339Nano`?

```
{{ colorGreen (rfc3339Nano $msg.ts) }}
```

cmd/cmd.go
niamster: Thank you for the suggestion, I didn't realise that `UseNumber` is not a default option!

cmd/cmd.go
niamster: Sounds good to me, I don't have a strong opinion here 👍

cmd/cmd.go
niamster: Definitely, thank you for pointing it out 🙇

cmd/cmd.go
niamster: Good point, I wanted to add that one overrides the other but that slipped from my mind.

niamster:
@tksm @superbrothers Thank you so much for the review!
I tried to address your comments, please let me know if I missed something.

NOTE: I added `toUTC` function to simplify testing. It still might be useful in some case I think.

cmd/cmd.go
tksm: Thanks for changing! Could you add the functions such as `toRFC3339Nano` to the table in the templates section?


image

README.md
```diff
@@ -247,6 +248,18 @@ Output using a custom template with `parseJSON`:
 stern --template='{{.PodName}}/{{.ContainerName}} {{with $d := .Message | parseJSON}}[{{$d.level}}] {{$d.message}}{{end}}{{"\n"}}' backend
 ```
 
+Output using a custom template that tries to parse JSON or fallbacks to raw format:
+
+```
+stern --template='{{.PodName}}/{{.ContainerName}} {{ with $msg := .Message | tryParseJSON }}[{{ colorGreen (toRFC3339Nano (toUTC $msg.ts)) }}] {{ levelColor $msg.level }} ({{ colorCyan $msg.caller }}) {{ $msg.msg }}{{ else }} {{ .Message }} {{ end }}{{"\n"}}' backend
```
tksm, line 254:
```suggestion
stern --template='{{.PodName}}/{{.ContainerName}} {{ with $msg := .Message | tryParseJSON }}[{{ colorGreen (toRFC3339Nano $msg.ts) }}] {{ levelColor $msg.level }} ({{ colorCyan $msg.caller }}) {{ $msg.msg }}{{ else }} {{ .Message }} {{ end }}{{"\n"}}' backend
```

Though `toUTC` is helpful for testing, using local time for readability might be better here.

tksm:
Thank you for the changes! I added minor comments below, but it looks good to me other than these. Could you rebase it to allow CI to check?

- https://github.com/stern/stern/pull/239#discussion_r1125155115
- https://github.com/stern/stern/pull/239#discussion_r1125160096

cmd/cmd.go
niamster: 👍 adding all of them them

niamster:
> Thank you for the changes! I added minor comments below, but it looks good to me other than these. Could you rebase it to allow CI to check?
> 
> * [allow flexible log parsing and formatting #239 (comment)](https://github.com/stern/stern/pull/239#discussion_r1125155115)
> * [allow flexible log parsing and formatting #239 (comment)](https://github.com/stern/stern/pull/239#discussion_r1125160096)

Thank you so for the suggestions! I merged `master` into my branch. D you want me to rebase instead (TBH I'm not a huge fan of rebase since it messes up with GH quite often, so we loose essential traces)?

tksm:
> Thank you so for the suggestions! I merged master into my branch. D you want me to rebase instead (TBH I'm not a huge fan of rebase since it messes up with GH quite often, so we loose essential traces)?

Thank you. Either rebase or merge is fine with me.

tksm (approved):
LGTM 👍

PR #253, opened 2023-03-31
Make setup-go get Go version from go.mod

tksm:
This PR makes setup-go [get the Go version from go.mod](https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file). It will eliminate the possibility of forgetting to update the Go version in the workflow files.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,7 @@ jobs:
     - uses: actions/checkout@v3
     - uses: actions/setup-go@v3
       with:
-        go-version: 1.19.x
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
     - uses: actions/checkout@v3
     - uses: actions/setup-go@v3
       with:
-        go-version: 1.19.x
+        go-version-file: go.mod
```
superbrothers, line 19: I did not know about this useful feature.

superbrothers (approved):

PR #254, opened 2023-04-05
Add support for the config file

### LINKED ISSUE #252: Add support for config file

Config file for stern.

**Why is this needed**:

If users want to use a non-default value for a flag in stern, the flag must be specified each time, which is inconvenient for daily operations. If users could arbitrarily change the default value of flags, this problem would be solved.

superbrothers:
This PR adds support for the config file.

This feature allows you to change the default values of stern options with the config file. The default config file path is `~/.config/stern/config.yaml`. 

```yaml
# <flag name>: <value>
tail: 10
max-log-requests: 999
timestamps: short
```

You can change the config file path with `--config` flag or `STERNCONFIG` environment variable.

Closes #252


README.md
```diff
@@ -109,6 +110,35 @@ Stern will use the `$KUBECONFIG` environment variable if set. If both the
...
+- `only-log-lines`
+- `output`
+- `selector`
+- `since`
+- `tail`
+- `template`
+- `template-file`
+- `timestamps`
+- `timezone`
+- `verbosity`
```
superbrothers, line 139: I am wondering if I should put them side by side.

> `color`, `container-state`, `ephemeral-containers`, `exclude`, `exclude-container`, `exclude-pod`, `field-selector`, `include`, `init-containers`, `max-log-requests`, `namespace`, `only-log-lines`, `output`, `selector`, `since`, `tail`, `template`, `template-file`, `timestamps`, `timezone`, `verbosity`

config.yaml
rkmathi, line 1: Is this file needed?

superbrothers: Yes, I added it by mistake.

rkmathi (approved):
LGTM🤩

rkmathi:
Sorry I had overlooked that this PR was in draft state.
I will check when draft state is over.

superbrothers:
> I will check when draft state is over.

@rkmathi  Now this PR is ready :pray:

cmd/cmd.go
```diff
@@ -39,6 +40,8 @@ import (
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 )
 
+const defaultConfigFilePath = "~/.config/stern/config.yaml"
```
tksm, line 43: [The tilde expansion does not work](https://stackoverflow.com/questions/17609732/expand-tilde-to-home-directory) without the shell, so we might need to use [os.UserHomeDir()](https://pkg.go.dev/os#UserHomeDir).

superbrothers: Oh... I'm going to use go-homedir `Expand`, since there seems to be a problem with "~" in `--config` flag and `STERNCONFIG` env.

tksm: The shell expands the tilde before calling a program, so I believe we do not need to implement it for flags and environment variables.

For example, `--config ~/stern.yaml` will be expanded to like `--config /home/ubuntu/stern.yaml` by the shell. `export STERNCONFIG=~/stern.yaml` will also be expanded in the same way.

```
$ export STERNCONFIG=~/stern.yaml
$ echo $STERNCONFIG
/home/ubuntu/stern.yaml
```

superbrothers: OK, I'll change to use `os.UserHomeDIr()`.

cmd/cmd.go
```diff
@@ -289,39 +305,77 @@ func (o *options) setVerbosity() error {
...
+	o.cf.Bind(fs.Lookup("template-file"))
 	fs.StringVarP(&o.timestamps, "timestamps", "t", o.timestamps, "Print timestamps with the specified format. One of 'default' or 'short'. If specified but without value, 'default' is used.")
+	o.cf.Bind(fs.Lookup("timestamps"))
 	fs.StringVar(&o.timezone, "timezone", o.timezone, "Set timestamps to specific timezone.")
+	o.cf.Bind(fs.Lookup("timezone"))
 	fs.BoolVar(&o.onlyLogLines, "only-log-lines", o.onlyLogLines, "Print only log lines")
+	o.cf.Bind(fs.Lookup("only-log-lines"))
+	fs.StringVar(&o.configFilePath, "config", o.configFilePath, "Path to the stern config file")
 	fs.IntVar(&o.verbosity, "verbosity", o.verbosity, "Number of the log level verbosity")
+	o.cf.Bind(fs.Lookup("verbosity"))
```
tksm, line 378: What do you think about making all flags available in the config file? This would simplify the logic and eliminate the need to maintain a list of available flags. Although the `version` and `completion` flags may not function properly in the config file, I don't think it would be a significant issue.

For example, the code would look something like the following.

```diff
diff --git a/cmd/cmd.go b/cmd/cmd.go
index 4060727..69066da 100644
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -19,7 +19,6 @@ import (
 	"encoding/json"
 	goflag "flag"
 	"fmt"
-	"io"
 	"os"
 	"regexp"
 	"strconv"
@@ -27,6 +26,7 @@ import (
 	"text/template"
 	"time"
 
+	"gopkg.in/yaml.v3"
 	"k8s.io/klog/v2"
 
 	"github.com/fatih/color"
@@ -79,7 +79,6 @@ type options struct {
 	node                string
 	configFilePath      string
 
-	cf                    *configFile
 	defaultConfigFilePath string
 }
 
@@ -104,7 +103,6 @@ func NewOptions(streams genericclioptions.IOStreams) *options {
 		maxLogRequests:      -1,
 		configFilePath:      defaultConfigFilePath,
 
-		cf:                    newConfigFile(),
 		defaultConfigFilePath: defaultConfigFilePath,
 	}
 }
@@ -305,77 +303,79 @@ func (o *options) setVerbosity() error {
 	return nil
 }
 
-func (o *options) ConfigFile() *configFile {
-	return o.cf
-}
-
-func (o *options) openConfigfile() (io.Reader, error) {
-	// empty data in the yaml format will be returned if the file does
-	// not exist in the default config file path.
+func (o *options) overrideFlagSetDefaultFromConfig(fs *pflag.FlagSet) error {
 	if o.configFilePath == o.defaultConfigFilePath {
 		if _, err := os.Stat(o.configFilePath); os.IsNotExist(err) {
-			return strings.NewReader("---\n"), nil
+			return nil
+		} else if err != nil {
+			return err
 		}
 	}
+	configFile, err := os.Open(o.configFilePath)
+	if err != nil {
+		return err
+	}
+
+	data := make(map[string]interface{})
+
+	if err := yaml.NewDecoder(configFile).Decode(data); err != nil {
+		return err
+	}
+
+	for name, value := range data {
+		flag := fs.Lookup(name)
+		fmt.Println(flag)
+		if flag == nil {
+			return fmt.Errorf("%q is not supported in the config file", name)
+		}
 
-	return os.Open(o.configFilePath)
+		// flag has higher priority than the config file
+		if flag.Changed {
+			continue
+		}
+
+		if err := flag.Value.Set(fmt.Sprint(value)); err != nil {
+			return fmt.Errorf("invalid value %q for %q in the config file: %v", value, flag.Name, err)
+		}
+	}
+
+	return nil
 }
 
 // AddFlags adds all the flags used by stern.
 func (o *options) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVarP(&o.allNamespaces, "all-namespaces", "A", o.allNamespaces, "If present, tail across all namespaces. A specific namespace is ignored even if specified with --namespace.")
 	fs.StringVar(&o.color, "color", o.color, "Force set color output. 'auto':  colorize if tty attached, 'always': always colorize, 'never': never colorize.")
-	o.cf.Bind(fs.Lookup("color"))
 	fs.StringVar(&o.completion, "completion", o.completion, "Output stern command-line completion code for the specified shell. Can be 'bash', 'zsh' or 'fish'.")
 	fs.StringVarP(&o.container, "container", "c", o.container, "Container name when multiple containers in pod. (regular expression)")
 	fs.StringSliceVar(&o.containerStates, "container-state", o.containerStates, "Tail containers with state in running, waiting, terminated, or all. 'all' matches all container states. To specify multiple states, repeat this or set comma-separated value.")
-	o.cf.Bind(fs.Lookup("container-state"))
 	fs.StringVar(&o.context, "context", o.context, "Kubernetes context to use. Default to current context configured in kubeconfig.")
 	fs.StringArrayVarP(&o.exclude, "exclude", "e", o.exclude, "Log lines to exclude. (regular expression)")
-	o.cf.Bind(fs.Lookup("exclude"))
 	fs.StringArrayVarP(&o.excludeContainer, "exclude-container", "E", o.excludeContainer, "Container name to exclude when multiple containers in pod. (regular expression)")
-	o.cf.Bind(fs.Lookup("exclude-container"))
 	fs.StringArrayVar(&o.excludePod, "exclude-pod", o.excludePod, "Pod name to exclude. (regular expression)")
-	o.cf.Bind(fs.Lookup("exclude-pod"))
 	fs.BoolVar(&o.noFollow, "no-follow", o.noFollow, "Exit when all logs have been shown.")
 	fs.StringArrayVarP(&o.include, "include", "i", o.include, "Log lines to include. (regular expression)")
-	o.cf.Bind(fs.Lookup("include"))
 	fs.BoolVar(&o.initContainers, "init-containers", o.initContainers, "Include or exclude init containers.")
-	o.cf.Bind(fs.Lookup("init-containers"))
 	fs.BoolVar(&o.ephemeralContainers, "ephemeral-containers", o.ephemeralContainers, "Include or exclude ephemeral containers.")
-	o.cf.Bind(fs.Lookup("ephemeral-containers"))
 	fs.StringVar(&o.kubeConfig, "kubeconfig", o.kubeConfig, "Path to kubeconfig file to use. Default to KUBECONFIG variable then ~/.kube/config path.")
 	fs.StringVar(&o.kubeConfig, "kube-config", o.kubeConfig, "Path to kubeconfig file to use.")
 	_ = fs.MarkDeprecated("kube-config", "Use --kubeconfig instead.")
 	fs.StringSliceVarP(&o.namespaces, "namespace", "n", o.namespaces, "Kubernetes namespace to use. Default to namespace configured in kubernetes context. To specify multiple namespaces, repeat this or set comma-separated value.")
-	o.cf.Bind(fs.Lookup("namespace"))
 	fs.StringVar(&o.node, "node", o.node, "Node name to filter on.")
 	fs.IntVar(&o.maxLogRequests, "max-log-requests", o.maxLogRequests, "Maximum number of concurrent logs to request. Defaults to 50, but 5 when specifying --no-follow")
-	o.cf.Bind(fs.Lookup("max-log-requests"))
 	fs.StringVarP(&o.output, "output", "o", o.output, "Specify predefined template. Currently support: [default, raw, json, extjson, ppextjson]")
-	o.cf.Bind(fs.Lookup("output"))
 	fs.BoolVarP(&o.prompt, "prompt", "p", o.prompt, "Toggle interactive prompt for selecting 'app.kubernetes.io/instance' label values.")
 	fs.StringVarP(&o.selector, "selector", "l", o.selector, "Selector (label query) to filter on. If present, default to \".*\" for the pod-query.")
-	o.cf.Bind(fs.Lookup("selector"))
 	fs.StringVar(&o.fieldSelector, "field-selector", o.fieldSelector, "Selector (field query) to filter on. If present, default to \".*\" for the pod-query.")
-	o.cf.Bind(fs.Lookup("field-selector"))
 	fs.DurationVarP(&o.since, "since", "s", o.since, "Return logs newer than a relative duration like 5s, 2m, or 3h.")
-	o.cf.Bind(fs.Lookup("since"))
 	fs.Int64Var(&o.tail, "tail", o.tail, "The number of lines from the end of the logs to show. Defaults to -1, showing all logs.")
-	o.cf.Bind(fs.Lookup("tail"))
 	fs.StringVar(&o.template, "template", o.template, "Template to use for log lines, leave empty to use --output flag.")
-	o.cf.Bind(fs.Lookup("template"))
 	fs.StringVarP(&o.templateFile, "template-file", "T", o.templateFile, "Path to template to use for log lines, leave empty to use --output flag. It overrides --template option.")
-	o.cf.Bind(fs.Lookup("template-file"))
 	fs.StringVarP(&o.timestamps, "timestamps", "t", o.timestamps, "Print timestamps with the specified format. One of 'default' or 'short'. If specified but without value, 'default' is used.")
-	o.cf.Bind(fs.Lookup("timestamps"))
 	fs.StringVar(&o.timezone, "timezone", o.timezone, "Set timestamps to specific timezone.")
-	o.cf.Bind(fs.Lookup("timezone"))
 	fs.BoolVar(&o.onlyLogLines, "only-log-lines", o.onlyLogLines, "Print only log lines")
-	o.cf.Bind(fs.Lookup("only-log-lines"))
 	fs.StringVar(&o.configFilePath, "config", o.configFilePath, "Path to the stern config file")
 	fs.IntVar(&o.verbosity, "verbosity", o.verbosity, "Number of the log level verbosity")
-	o.cf.Bind(fs.Lookup("verbosity"))
 	fs.BoolVarP(&o.version, "version", "v", o.version, "Print the version and exit.")
 
 	fs.Lookup("timestamps").NoOptDefVal = "default"
@@ -540,13 +540,7 @@ func NewSternCmd(stream genericclioptions.IOStreams) (*cobra.Command, error) {
 				return err
 			}
 
-			r, err := o.openConfigfile()
-			if err != nil {
-				return err
-			}
-
-			// Load options from the config file
-			if err := o.cf.LoadFromReader(r); err != nil {
+			if err := o.overrideFlagSetDefaultFromConfig(cmd.Flags()); err != nil {
 				return err
 			}
```

```go
func (o *options) overrideFlagSetDefaultFromConfig(fs *pflag.FlagSet) error {
	if o.configFilePath == o.defaultConfigFilePath {
		if _, err := os.Stat(o.configFilePath); os.IsNotExist(err) {
			return nil
		} else if err != nil {
			return err
		}
	}
	configFile, err := os.Open(o.configFilePath)
	if err != nil {
		return err
	}

	data := make(map[string]interface{})

	if err := yaml.NewDecoder(configFile).Decode(data); err != nil {
		return err
	}

	for name, value := range data {
		flag := fs.Lookup(name)
		fmt.Println(flag)
		if flag == nil {
			return fmt.Errorf("%q is not supported in the config file", name)
		}

		// flag has higher priority than the config file
		if flag.Changed {
			continue
		}

		if err := flag.Value.Set(fmt.Sprint(value)); err != nil {
			return fmt.Errorf("invalid value %q for %q in the config file: %v", value, flag.Name, err)
		}
	}

	return nil
}
```

superbrothers: Thanks for the suggestion. I agree. We can keep the code simple by not making explicit the options available in the configuration file.

Due to a major rewrite, this PR will be changed to Draft once.

superbrothers:
@tksm I've updated this PR. Please review it again.

cmd/cmd.go
```diff
@@ -107,6 +113,11 @@ func (o *options) Complete(args []string) error {
 		}
 	}
 
+	envVar, ok := os.LookupEnv("STERNCONFIG")
+	if ok {
+		o.configFilePath = envVar
+	}
```
tksm, line 119: `Complete()` is called after `overrideFlagSetDefaultFromConfig()`, so it may not work as expected.

```
$ STERNCONFIG=$PWD/cmd/testdata/config-invalid.yaml dist/stern .
# It should fail, but we currently get no error.
```

superbrothers: Fixed. e00a0e1

tksm (approved):
Great feature! LGTM 🎉

rkmathi:
lgtm! 😸 :

PR #255, opened 2023-04-10
Allow an empty config file

tksm:
This PR allows an empty config file. Currently, stern fails as follows when the config file is empty. Please see https://github.com/go-yaml/yaml/issues/639 for details.

```
$ touch /tmp/empty-sternconfig.yaml

$ dist/stern --config /tmp/empty-sternconfig.yaml .
Error: EOF
Usage:
  stern pod-query [flags]
# ...
```

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	goflag "flag"
 	"fmt"
+	"io"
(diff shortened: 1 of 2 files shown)


rkmathi (approved):
LGTM 👍

superbrothers (approved):

PR #260, opened 2023-04-14
Integrate fmt and vet checks into golangci-lint

tksm:
This PR integrates fmt and vet checks in Makefile into golangci-lint. It configures golangci-lint to enable [gofmt](https://golangci-lint.run/usage/linters/#gofmt) and [the default linters, including govet](https://golangci-lint.run/usage/linters/#enabled-by-default).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,13 @@
+run:
+  timeout: 5m
+linters:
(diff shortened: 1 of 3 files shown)


superbrothers (approved):
Nice!

Issue #263, opened 2023-05-03
Missing log level keys and fallback


nomme:
**What happened**:
When using `--template` with the function `levelColor` it prints this error, without the log entry, when encountering level keys it's not familiar with.
```
expanding template failed: template: log:1:148: executing "log" at <levelColor $d.levelname>: error calling levelColor: runtime error: invalid memory address or nil pointer dereference
```

**What you expected to happen**:
I expect the most common level keys to be handled and a fallback to exist instead of generating an error.

**How to reproduce it (as minimally and precisely as possible)**:
Produce "warning" and "critical" logs from a Python application. These two levelnames are not handled.

If stern could read from stdin (I couldn't find that it does at least) it could have looked something like this.
```sh
echo '{"levelname": "warning"}' | stern --template='{{with $d := .Message | tryParseJSON}}{{levelColor $d.levelname}}{{end}}{{"\n"}}'
```

**Anything else we need to know?**:
I'm submitting a PR for this in a moment. 

**Environment**:
- stern version (use `stern --version`): 
```
version: 1.24.0
commit: 68305a0ab1383053166b7b606ad167b43fc6b329
built at: 2023-03-12T06:50:44Z
```
- OS (e.g: `cat /etc/os-release`): Arch Linux
- Install tools (e.g: Homebrew): Arch AUR https://aur.archlinux.org/packages/stern-bin

nomme:
PR merged, closing

PR #264, opened 2023-05-03
Add log level color keys and handle default

nomme:
This PR intends to solve issue #263 by handling two more common log-level keys as well as handling the fallback case so that an error is never printed. The two new keys are used in Python's logging facilities.

I didn't find a way to add custom log-level keys as a user, but I think this works well. It would be very appreciated if it didn't cause an error at least.

Test by using stern to print logs from a Python application that emits "warning" and "critical" logs. Preferably also test with an unknown key 
```sh
echo '{"levelname": "invalid key"}' | stern --template='{{with $d := .Message | tryParseJSON}}{{levelColor $d.levelname}}{{end}}{{"\n"}}'
```

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -499,7 +499,7 @@ func (o *options) generateTemplate() (*template.Template, error) {
 			case "critical":
 				levelColor = color.New(color.FgCyan)
 			default:
-				levelColor = color.New(color.FgWhite)
+				return level
 			}
 			return levelColor.SprintFunc()(level)
 		},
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -494,7 +496,10 @@ func (o *options) generateTemplate() (*template.Template, error) {
 				levelColor = color.New(color.FgRed)
 			case "fatal":
 				levelColor = color.New(color.FgCyan)
+			case "critical":
+				levelColor = color.New(color.FgCyan)
 			default:
+				levelColor = color.New(color.FgWhite)
```
tksm, line 502:
```suggestion
				return level
```

Since some terminals have a default background color of white, maybe we should avoid using white. How about returning text with the default color?

nomme: Yeah, that's a great idea!

tksm (approved):
LGTM :+1: Thanks for your PR!

PR #276, opened 2023-08-25
Add `--condition`

### LINKED ISSUE #213: `--exclude-healthy-pods`
This is somewhat similar to https://github.com/stern/stern/issues/189, but this isn't about making stern exit when reaching a condition.

Basically, I only want stern to show the logs for the pods which aren't in `healthy` state. As soon as the pod becomes `healthy`, stern should stop capturing the logs for that pod as well.

felipecrs:
Example:

```console
stern . --condition=ready=false --tail=0
```

TODOs:

- [x] Only starts tailing logs for pods matching --condition
- [x] Stop tailing logs for pods that no longer match --condition
- [x] Add tests

Fixes #213


felipecrs:
I'm giving #213 because I really need it. But I'm struggling with the second TODO. Is there a chance someone can help me?

PS: code is not yet ready for final review, it's just working. Also, I have almost zero experience with Golang.

superbrothers:
I will look into it this weekend.

felipecrs:
Thanks a lot @superbrothers!

superbrothers:
It is on the way, but I have a working patch. I won't have much time this weekend but will be able to work on it after that. https://github.com/stern/stern/commit/816c8960d68e14642238230bb65fe588039e7e16

You can try it out as soon as you build it.

felipecrs:
@superbrothers, thanks a lot, you are the best!

However, it's not working quite as I expected. Let me help build a testing scenario too:

```yaml
kubectl apply -f - <<'EOF'
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test
  serviceName: test
  template:
    metadata:
      labels:
        app: test
    spec:
      containers:
      - name: test
        image: bash
        command:
          - bash
          - -c
          - |
            count=0
            while true; do
              echo "Hello World: $count"
              count=$((count+1))
              if [[ $count -eq 10 ]]; then
                touch /tmp/healthy
              fi
              sleep 1
            done
        readinessProbe:
          exec:
            command:
            - cat
            - /tmp/healthy
          initialDelaySeconds: 10
          periodSeconds: 1
EOF
```

This pod becomes ready in about 11 seconds.

Start the stern process with:

```console
./dist/stern '.*' --condition ready=false --tail 0
```

Stern correctly filters out the pod if the pod is already ready, but Stern does not stop tailing the logs when the pod becomes ready:

https://github.com/stern/stern/assets/29582865/fdd4a9cd-1463-4733-911b-fe5ae6b0935e

felipecrs:
Ok, I think I know what's going on: K8s does not seem to emit an event for when the pod becomes ready. I'm investigating what can be done about it.

tksm:
Hi,

It appears that Stern has detected the events, but they could be filtered out by [targetFilter.shouldAdd()](https://github.com/superbrothers/stern/blob/816c8960d68e14642238230bb65fe588039e7e16/stern/target.go#L139-L141).

The following patch may resolve the issue, but I've noticed the behavior of `--condition ready=false` can be somewhat confusing. When we delete pods, the pod's `Ready` status reverts to `False`, causing the logs to restart from the beginning. Additionally, the `Ready` status can change due to other reasons.

```diff
diff --git a/stern/stern.go b/stern/stern.go
index c492519..dcece50 100644
--- a/stern/stern.go
+++ b/stern/stern.go
@@ -222,7 +222,7 @@ func Run(ctx context.Context, config *Config) error {
 					if !ok {
 						continue
 					}
-					cancel.(func())()
+					cancel.(context.CancelFunc)()
 				case <-nctx.Done():
 					return nil
 				}
diff --git a/stern/target.go b/stern/target.go
index 553e8c2..cbc6788 100644
--- a/stern/target.go
+++ b/stern/target.go
@@ -136,8 +136,14 @@ OUTER:
 			Container: c.Name,
 		}
 
+		if !conditionFound {
+			visitor(t, false)
+			f.forget(string(pod.UID))
+			continue
+		}
+
 		if f.shouldAdd(t, string(pod.UID), c) {
-			visitor(t, conditionFound)
+			visitor(t, true)
 		}
 	}
 }
```

The demo below illustrates that the logs restart when the pod is deleted (at 00:18).

[asciicast](https://asciinema.org/a/PsymNLbipgbpedYmFfLT2aXr9)

felipecrs:
Wow, it looks to be working indeed as per your demo. Can't wait to try your patch out.

BTW the behavior you described is indeed misleading.

That would not happen though when using --tail=0, which is my use case (only capture live logs, not past logs).

I will think about what can be done to circumvent this.

Maybe Stern has to remember up until which point it tailed logs for a given pod, and when it starts tailing it again, it should tail from that point on.

tksm:
@felipecrs Oh, sorry. I missed the `--tail=0` option. Thank you for pointing it out.

As you mentioned, Stern behaves as expected when using `--tail=0`.

[asciicast](https://asciinema.org/a/tKbDvkByBZgUpDSq9B9yPPOid)

felipecrs:
@tksm really amazing, indeed it works like a charm. I added both @superbrothers' commit and your patch to this PR now.

Something I believe I need to do before merging is to fix the condition syntax for the config file. I left it as a string there, which makes it exactly as the CLI, but perhaps I should convert it to an object like:

```yaml
- condition: ready=false
+ condition:
+   name: ready
+   value: false
```

What do you think? Personally, I don't mind. Even the first one looks good to me.

Another thing to think about is the concern raised by @tksm above when not using `--tail=0`. I proposed one solution above but I'm not sure if it's worth implementing it. Another option is to simply have this caveat acknowledged somewhere, maybe in the README.

tksm:
I'm glad to hear it worked as expected. :tada:

#### Config file
I believe that the first one (`condition: ready=false`) is better because it maintains consistency with other options. Additionally, we do not need to implement anything in that case.

#### --tail=0 option
For the initial implementation, I think that the caveat in README.md and the error message when `--tail=0` is not set are sufficient. Maybe we should avoid automatically setting `--tail=0` for future compatibility.

```bash
# Raise an error when `--tail=0` is not set
$ ./dist/stern '.*' --condition ready=false
Error: Currently --condition and --no-follow=false only works with --tail=0
Usage:
  stern pod-query [flags]
```

```diff
diff --git a/cmd/cmd.go b/cmd/cmd.go
index 35e3060..2c70027 100644
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -130,6 +130,9 @@ func (o *options) Validate() error {
 	if o.selector != "" && o.resource != "" {
 		return errors.New("--selector and the <resource>/<name> query can not be set at the same time")
 	}
+	if o.condition != "" && !o.noFollow && o.tail != 0 {
+		return errors.New("Currently --condition and --no-follow=false only works with --tail=0")
+	}
 
 	return nil
 }
```

felipecrs:
I'm sorry @tksm, but why enforce `--tail=0` to `--no-follow=false`? Maybe you meant `--no-follow=true`, where it makes more sense: if `--no-follow=true`, no live logs would be shown, only past logs. Thus, filtering only live logs with `--tail=0` makes no sense.

I changed it in the commit.

felipecrs:
I also just realized that `--condition=ready=false` will filter out pods which does not have a health check.

I need to implement something more flexible, like `--condition=ready!=true`.

felipecrs:
Another thing, @tksm, I think allowing people to specify `--tail` more than `0` can be useful. Imagine a situation like:

_I want to only display logs for pods which are not ready, but displaying their last 3 lines before they became unhealthy_

felipecrs:
Ok, now it fulfils my use case:

```
stern . --condition=ready=false --tail=0 --only-condition-pods-with-readiness
```

Please let me know what you think.

felipecrs:
Ok, I have been soaking this internally during the day and the results are really good for my use case.

I would like to ask the maintainers to give me the go ahead on the current solution (`--only-condition-pods-with-readiness`) before I start cleaning up the PR and working on fixing the tests.

tksm:
@felipecrs Let me clarify the conditions under which you want to view pod logs in your use case.

I assume you do **NOT** want to view logs when:

1. When the pod has at least one readinessProbe **OR** ReadinessGate.
1. **AND** the `pod.conditions[]` has `Ready=True`.

Please inform me if there are any additional conditions.

In that case, I believe specifying these conditions using flags might be overly complex.

I came up with another approach with `--template`.

It may be possible to address this use case using the `--template` by adding the reference to the latest pod as a template variable and a helper function like `podIsReady` instead of adding the flags. What are your thoughts on this approach?

Although templates can be complex, this approach offers the flexibility to handle a wide range of use cases. You can use `--template-file` if you want to store the template in a separate file.

```
./dist/stern . --template '
{{- $pod := .Reference.GetPod -}}
{{- $hasReadinessGates := gt (len $pod.Spec.ReadinessGates) 0 -}}
{{- $hasReadinessProbe := false}}
{{- range $c := $pod.Spec.Containers -}}
  {{- if $c.ReadinessProbe -}}
    {{- $hasReadinessProbe = true -}}
  {{- end -}}
{{- end -}}
{{- if not (and (or $hasReadinessProbe $hasReadinessGates) (podIsReady $pod)) -}}
{{- .Message -}}{{"\n"}}
{{- end -}}
'
```

felipecrs:
Isn't that a very costly operation? Checking the pod per log line as opposed to only per event?

felipecrs:
But yes, this set of conditions satisfies my use case, which can be described in high level with something like:

I want Stern to tail logs from all pods which are not healthy.

If the pod does not have a health checker (readiness probe), then I want to tail its logs since I can't tell whether it's healthy or not.

Precisely, I want to run Stern as a side car process to my helm upgrade command within my CI/CD pipeline, and these conditions allows for tracking the helm installation as it happens more decently than just showing all pod logs no matter what.

In my use case, I have something around 50 pods to tail, but I install them incrementally (one helm chart per time). Filtering out pods which are healthy ensures I'm not showing logs for pods which were installed by previous helm installations.

That said, I believe more people could leverage the same approach.

If you really believe we should not be too narrow in the implementation, maybe we can invent some kind of --condition-template, which allows to run a template that should result in a "false" or "true" to decide whether the pod should be tailed or not. And in that case, I believe not even the helper podIsReady function needs to exist, as it can be easily implemented in the template itself.

tksm:
Thanks for your detailed explanation!

I understand the use case, and it sounds worth implementing. @superbrothers What do you think?

I prefer creating a dedicated flag for this use case to improve clarity, but I find the term `unhealthy` in `--only-unhealthy-pods` somewhat ambiguous. How about using a more specific name like `--exclude-readiness-succeeded`.

felipecrs:
I don't mind going the `--condition-template` approach BTW. :)

tksm:
If you don't mind using a complex template, the `--condition-template` approach looks better since it can be used in a wide range of use cases.

For your reference, if you would have the `pod` variable in a condition template, the template for your use case would be something like below.

```
{{- $podIsReady := false -}}
{{- range $c := $pod.Status.Conditions -}}
  {{- if and (eq $c.Type "Ready") (eq $c.Status "True") -}}
    {{- $podIsReady = true -}}
  {{- end -}}
{{- end -}}

{{- $hasReadinessProbe := false}}
{{- range $c := $pod.Spec.Containers -}}
  {{- if $c.ReadinessProbe -}}
    {{- $hasReadinessProbe = true -}}
  {{- end -}}
{{- end -}}

{{- $hasReadinessGates := gt (len $pod.Spec.ReadinessGates) 0 -}}

{{- not (and (or $hasReadinessProbe $hasReadinessGates) (podIsReady $pod))}}
```

It would be great if it had a debug functionality, such as the ability to print a rendered template. For example, I debug my template by showing variables like below when using `--template`.

```
{{- if $skip -}}
[skip] {{.Message}}
{{- else -}}
{{colorGreen .Message}}
{{- end }} / ready={{$podIsReady}}, probe={{$hasReadinessProbe}}, gates={{$hasReadinessGates}}
{{- "\n" -}}
```

image

guettli:
btw, I wrote a small tool to check all conditions in a cluster: https://github.com/guettli/check-conditions PRs are welcome.

felipecrs:
@guettli, do you mean stern could use your tool as a library to handle this condition-thing?

felipecrs:
@tksm, I have simplified and completed this PR. I believe it's good to be merged. Can you please check it again?

felipecrs:
@superbrothers feel free to take a look at this one too. 😅

felipecrs:
Got it. Thank you. For now the current implementation of this PR should be good enough, but I'll keep an eye on it.

tksm (commented):
Hi, thanks for the implementation! :tada:

I've confirmed that it works as intended with `--tail=0`. The manifest I used for testing is attached below.

However, I noticed that using `--condition` without `--tail=0` or `--no-follow` can lead to somewhat confusing behavior when the condition changes repeatedly. Would you consider adding a note to the documentation and help about this? Alternatively, setting `--tail=0` as the default when `--condition` is specified might help clarify things.

Another option is to automatically set `TailLines=0` when the pod's log has already been displayed once, ensuring that logs start from the next event.

### Confirmation
stern-condition


These are three outputs I confirmed with different options.

1. without `--condition`: all logs from the pod are shown
2. with `--condition=ready`: only logs are shown when the pod's condition is ready
3. with `--condition=ready=false`: only logs are shown when the pod's condition is not ready

### Manifest

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: ready-flipper
spec:
  containers:
  - name: ready-flipper
    image: bash
    command:
      - bash
      - -c
      - |
        for _ in $(seq 2); do
          rm -f /tmp/ready; sleep 3
          for _ in $(seq 3); do echo not ready; sleep 1; done
          touch /tmp/ready; sleep 3
          for _ in $(seq 3); do echo ready; sleep 1; done
        done
        sleep 1000000
    readinessProbe:
      exec:
        command:
        - cat
        - /tmp/ready
      initialDelaySeconds: 1
      periodSeconds: 1
      failureThreshold: 1
```

stern/stern.go
```diff
@@ -197,10 +200,24 @@ func Run(ctx context.Context, client kubernetes.Interface, config *Config) error
...
+					ctx, cancel := context.WithCancel(context.Background())
+					m.Store(target.GetID(), cancel)
 					go func() {
-						tailTarget(nctx, target)
+						go tailTarget(ctx, target)
+						select {
+						case <-ctx.Done():
+						case <-nctx.Done():
+							cancel()
+						}
```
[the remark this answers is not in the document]
felipecrs: It's funny I no longer understand anything of the code I wrote. Allow me some time to handle this. :D

felipecrs:
> Another option is to automatically set TailLines=0 when the pod's log has already been displayed once, ensuring that logs start from the next event.

@tksm, this would be the best. I'll try to implement it.

tksm:
Perhaps this can be achieved by skipping the [f.forget()](https://github.com/felipecrs/stern/blob/19a5cec50f1d212ecdefc7cd35ed3ce2a72e3084/stern/target.go#L123) call and adding a new field to [targetFilter.targetStates](https://github.com/felipecrs/stern/blob/19a5cec50f1d212ecdefc7cd35ed3ce2a72e3084/stern/target.go#L48) to track whether the pod's log has already been displayed. However, it might be complex, so I'm not sure if it's worth implementing.
(Sorry, I don't have time to implement this myself.)

felipecrs:
@tksm thank you very much for the suggestion, but I was still not able to implement anything useful.

So, I just added a flag check for now. Perhaps we can improve this in the future, but I don't think it's worth hold this feature because of it.

cmd/cmd.go
```diff
@@ -161,11 +163,14 @@ func (o *options) Validate() error {
...
 	if o.selector != "" && o.resource != "" {
-		return errors.New("--selector and the <resource>/<name> query can not be set at the same time")
+		return errors.New("--selector and the <resource>/<name> query cannot be set at the same time")
 	}
 	if o.noFollow && o.tail == 0 {
 		return errors.New("--no-follow cannot be used with --tail=0")
 	}
+	if o.condition != "" && o.tail != 0 {
+		return errors.New("--condition is currently only supported with --tail=0")
+	}
```
tksm, line 173: How about excluding `--no-follow` from the validation, since it doesn't accept `--tail=0` but can work with the `--tail` option?

```suggestion
	if o.condition != "" && o.tail != 0 && !o.noFollow {
		return errors.New("--condition is currently only supported with --tail=0 or --no-follow")
	}
```

felipecrs: You're right. I made some tests and --no-follow indeed works ok with --condition.

cmd/cmd_test.go
```diff
@@ -134,6 +134,30 @@ func TestOptionsValidate(t *testing.T) {
...
+			}(),
+			"--condition is currently only supported with --tail=0 or --no-follow",
+		},
+		{
+			"Specify --condition without --no-follow and no --tail=0",
+			func() *options {
+				o := NewOptions(streams)
+				o.podQuery = "."
+				o.condition = "ready=false"
+				o.noFollow = true
```
tksm, line 155:
```suggestion
```

This test case currently seems to fail. Should this be removed, as it is described as `without --no-follow`?

```
$ go test github.com/stern/stern/cmd -run TestOptionsValidate
--- FAIL: TestOptionsValidate (0.00s)
    --- FAIL: TestOptionsValidate/Specify_--condition_without_--no-follow_and_no_--tail=0 (0.00s)
        cmd_test.go:208: expected "--condition is currently only supported with --tail=0 or --no-follow" err, but actual no err
FAIL
FAIL    github.com/stern/stern/cmd      0.986s
FAIL
```

felipecrs: ops.

tksm (approved):
LGTM :rocket:

Thank you very much for the contribution!

felipecrs:
Thank you for patiently helping me complete this PR!

PR #286, opened 2023-12-07
README.md - make text reference for pod-query match formatting of cli example

rouilj:
The cli example shows "`pod-query`". The text refers to "`pod` query". These two references should match. This patch does that.


tksm (approved):
LGTM 🚀

PR #289, opened 2023-12-28
adding a highlight option

prune998:
This PR adds a new `--highlight` (or `-I`) option that will use a regexp to highlight some parts of the logs. 

This is similat to the `--include` option behaviour but will not filter logs, only add color.

### DIFF
--- a/stern/tail.go
+++ b/stern/tail.go
@@ -72,6 +72,7 @@ type TailOptions struct {
 	SinceTime    *metav1.Time
 	Exclude      []*regexp.Regexp
 	Include      []*regexp.Regexp
+	Highlight    []*regexp.Regexp
 	Namespace    bool
 	TailLines    *int64
 	Follow       bool
@@ -113,14 +114,17 @@ func (o TailOptions) IsInclude(msg string) bool {
 var colorHighlight = color.New(color.FgRed, color.Bold).SprintFunc()
 
 func (o TailOptions) HighlightMatchedString(msg string) string {
-	if len(o.Include) == 0 {
+	if len(o.Include) == 0 && len(o.Highlight) == 0 {
 		return msg
 	}
 
 	if o.reHightlight == nil {
-		ss := make([]string, len(o.Include))
+		ss := make([]string, len(o.Include)+len(o.Highlight))
 		for i, rin := range o.Include {
 			ss[i] = rin.String()
+		}
+		for i, rhi := range o.Highlight {
+			ss[i] = rhi.String()
 		}
 
 		// We expect a longer match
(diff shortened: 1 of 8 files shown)


prune998:
Anyone ?

stern/tail.go
```diff
@@ -113,15 +114,18 @@ func (o TailOptions) IsInclude(msg string) bool {
...
 		return msg
 	}
 
 	if o.reHightlight == nil {
-		ss := make([]string, len(o.Include))
+		ss := make([]string, len(o.Include)+len(o.Highlight))
 		for i, rin := range o.Include {
 			ss[i] = rin.String()
 		}
+		for i, rhi := range o.Highlight {
```
tksm, line 126: It seems that it overwrites the values from --include, so it does not work correctly when both --include and --highlight are specified. How about merging the two slices first?

```diff
diff --git a/stern/tail.go b/stern/tail.go
index 8e9ce2c..17d23ac 100644
--- a/stern/tail.go
+++ b/stern/tail.go
@@ -114,17 +114,15 @@ func (o TailOptions) IsInclude(msg string) bool {
 var colorHighlight = color.New(color.FgRed, color.Bold).SprintFunc()
 
 func (o TailOptions) HighlightMatchedString(msg string) string {
-	if len(o.Include) == 0 && len(o.Highlight) == 0 {
+	highlights := append(o.Include, o.Highlight...)
+	if len(highlights) == 0 {
 		return msg
 	}
 
 	if o.reHightlight == nil {
-		ss := make([]string, len(o.Include)+len(o.Highlight))
-		for i, rin := range o.Include {
-			ss[i] = rin.String()
-		}
-		for i, rhi := range o.Highlight {
-			ss[i] = rhi.String()
+		ss := make([]string, len(highlights))
+		for i, hl := range highlights {
+			ss[i] = hl.String()
 		}
 
 		// We expect a longer match
```

tksm (changes requested):
Sorry for the late reply. It is a neat feature! I commented on a minor issue when both --include and --highlight specified.

stern/tail.go
prune998: Thanks for the review. I thought I tested it... seems not :)

I applied your changes + added some tests.

I set the highlight option to `-I` but maybe `-H` would be better ? Let me know if you want me to change that !

prune998: I went ahead and replaced the option by `-H`

tksm: Hi,

Thanks for the update! I agree that using `-H` is more intuitive. :+1:

tksm (approved):
LGTM 🚀

PR #300, opened 2024-05-02
Add support for UNIX time with nanoseconds to template functions

### LINKED ISSUE #297: `toUTC` unable to parse Kubernetes controller-runtime log timestamps
**What happened**:

Attempting to format the timestamp from the JSON structured logs of a Go-based Kubernetes controller (specifically external-secrets, but this shows the relevant format - https://kubernetes.io/blog/2020/09/04/kubernetes-1-19-introducing-structured-logs/#using-structured-logs) results in the incorrect date

The timestamp, which is seconds past epoch but with milliseconds (?) after the decimal point, to be parsed by `toUTC`  correctly

```shell
GOBIN=$(pwd) go install github.com/stern/stern@c1649ca9f88f9183467dadda7cd8c0ffd16943e8
echo '{"ts":1580306777.04728,"msg":"Pod status updated"}' | ./stern --stdin=true --template '{{- with $msg := .Message | tryParseJSON -}}{{- toUTC $msg.ts -}}{{- " - " -}}{{- $msg.msg -}}{{- end -}}'
```

The underlying bug appears to be in github.com/spf13/cast which converts the `json.Number` into a `time.Time` by turning into  a string before attempting to turn it into an `int64` , which it assumes to be a whole number. Unfortunately, that repository has had much activity since February 2023.

**Environment**:
- stern version (use `stern --version`): Built from HEAD (`c1649ca9f88f9183467dadda7cd8c0ffd16943e8`) to demo with `stdin` flag, but also present in 1.28.0
- OS (e.g: `cat /etc/os-release`): Both macOS and Ubuntu 23.10
- Install tools (e.g: Homebrew): Homebrew
- Others:

tksm:
This commit adds support for UNIX time seconds with nanoseconds (e.g., `1136171056.02`) to the following template functions:

- `toRFC3339Nano`
- `toTUC`
- `toTimestamp`

As described in #297, the `toUTC` currently fails to parse a timestamp like `1580306777.04728`, resulting in the incorrect date.

```
$ echo '{"ts":1580306777.04728,"msg":"Pod status updated"}' | ./stern --stdin=true --template '{{- with $msg := .Message | tryParseJSON -}}{{- toUTC $msg.ts -}}{{- " - " -}}{{- $msg.msg -}}{{- end -}}'
0001-01-01 00:00:00 +0000 UTC - Pod status updated
```

With this commit, the `toUTC` can correctly parse the timestamp with nanoseconds.

```
2020-01-29 14:06:17.04728 +0000 UTC - Pod status updated%
```

Fixes #297

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -30,7 +30,6 @@ import (
 	"github.com/fatih/color"
 	"github.com/mitchellh/go-homedir"
 	"github.com/pkg/errors"
-	"github.com/spf13/cast"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/stern/stern/stern"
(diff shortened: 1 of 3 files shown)


rkmathi (approved):
LGTM 🥳

Issue #304, opened 2024-05-11
Support print logs in different colors for each container


cr7258:
**What would you like to be added**:

Support print logs in different colors for each container, like [kubetail](https://github.com/johanhaleby/kubetail).
<img width="777" alt="image" src="https://github.com/stern/stern/assets/40051120/91c3188d-bec1-4c90-9cd1-c3defae3d6e2">

**Why is this needed**:
Improve the readability and clarity of multiple containers' logs.

PR #305, opened 2024-05-11
Display different colors for different containers

cr7258:
fixes https://github.com/stern/stern/issues/304
All containers within a pod have the same color by default.
```bash
./stern myapp
```
image

Set `--diff-container` to display different colors for different containers. This is useful when you want to debug logs for multiple containers in the same pod.

```bash
# --diff-container is false by default
./stern myapp --diff-container
```
image

### DIFF
--- a/stern/tail.go
+++ b/stern/tail.go
@@ -67,8 +67,8 @@ type ResumeRequest struct {
 }
 
 // NewTail returns a new tail for a Kubernetes container inside a pod
-func NewTail(clientset corev1client.CoreV1Interface, nodeName, namespace, podName, containerName string, tmpl *template.Template, out, errOut io.Writer, options *TailOptions) *Tail {
-	podColor, containerColor := determineColor(podName)
+func NewTail(clientset corev1client.CoreV1Interface, nodeName, namespace, podName, containerName string, tmpl *template.Template, out, errOut io.Writer, options *TailOptions, diffContainer bool) *Tail {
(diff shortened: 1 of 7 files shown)


stern/tail.go
```diff
@@ -96,13 +96,22 @@ var colorList = [][2]*color.Color{
...
+		_, _ = containerHash.Write([]byte(containerName))
+		containerIdx := containerHash.Sum32() % uint32(len(colorList))
+		containerColor = colorList[containerIdx][1]
+	} else {
+		containerColor = colors[1]
+	}
 
-	colors := colorList[idx]
-	return colors[0], colors[1]
+	return colors[0], containerColor
```
tksm, line 114: How about moving the index calculation into a separate function?

```go
func determineColor(podName, containerName string, diffContainer bool) (podColor, containerColor *color.Color) {
	colors := colorList[colorIndex(podName)]
	if diffContainer {
		return colors[0], colorList[colorIndex(containerName)][1]
	}
	return colors[0], colors[1]
}

func colorIndex(name string) uint32 {
	hash := fnv.New32()
	_, _ = hash.Write([]byte(name))
	return hash.Sum32() % uint32(len(colorList))
}
```

cr7258: @tksm Thanks, applied your suggestion.

tksm (approved):
LGTM :rocket:

:memo: We can also enable this feature in the config file (`~/.config/stern/config.yaml`).

```yaml
diff-container: true
```

PR #315, opened 2024-10-06
Update dependencies for Kubernetes 1.31

tksm:
This PR updates the tools and the dependencies for the release of Kubernetes v1.31.


.goreleaser.yaml
```diff
@@ -1,3 +1,4 @@
+version: 2
```
tksm, line 1: :memo: GoReleaser v2 requires the version 2 config format.

> It should only warn you about the version header in the configuration file, which you can fix by adding version: 2 to it.
https://goreleaser.com/blog/goreleaser-v2/

superbrothers (approved):
Thanks!

PR #321, opened 2024-11-09
Allow `levelColor` template function to parse numbers

### LINKED ISSUE #320: Colorize numeric log levels
Colorize numeric log levels with `levelColor` or similar, or at least allow the log to be printed anyway.
E.g. `{"level": 30, "msg": "A message"}`

**Why is this needed**:
Some log libraries use numeric log levels instead of the, in my experience, more usual debug, info, warn etc. When using the template function `levelColor` on numeric values it throws an error.
```
expanding template failed: template: log:1:70: executing "log" at <$d.level>: wrong type for value; expected string; got json.Number
```
It would be nice if `levelColor` could colorize numbers too, or at least allow the log line to be printed anyway.

hogklint:
This fixes #320

This PR extends the functionality of the template function `levelColor` to also support numeric values. Try it out with
```
for l in 10 20 30 40 50 60 70 \"debug\" \"info\" \"warn\" \"error\" \"fatal\"; do echo '{"level": '$l', "msg": "A message"}'; done | stern --stdin --template='{{with $d := .Message | parseJSON}}[{{levelColor $d.level}}] {{$d.msg}}{{end}}{{"\n"}}'
```

In case it cannot parse the value as a string or number it currently falls back to returning an empty string without an error. I think this is quite nice as it allows for things like this `{{or (colorLevel $d.level) "N/A"}}` if the level is missing completely.  It is however a behavior change.

Furthermore, it can be debated whether the levels I've chosen make the most sense. They are set according to what [bunyan](https://github.com/trentm/node-bunyan?tab=readme-ov-file#levels) and [pino](https://github.com/pinojs/pino/blob/main/docs/api.md#loggerlevels-object) uses whereas [log4j](https://logging.apache.org/log4j/2.x/manual/customloglevels.html) use levels in the 100s. A flag to switch between the two styles perhaps?

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,7 @@ functions](https://golang.org/pkg/text/template/#hdr-Functions)):
 | `toRFC3339Nano` | `object`              | Parse timestamp (string, int, json.Number) and output it using RFC3339Nano format |
 | `toTimestamp`   | `object, string [, string]` | Parse timestamp (string, int, json.Number) and output it using the given layout in the timezone that is optionally given (defaults to UTC). |
 | `levelColor`    | `string`              | Print log level using appropriate color                                           |
+| `bunyanLevelColor`    | `string`        | Print [bunyan](https://github.com/trentm/node-bunyan) numeric log level using appropriate color |
 | `colorBlack`    | `string`              | Print text using black color                                                      |
 | `colorRed`      | `string`              | Print text using red color                                                        |
 | `colorGreen`    | `string`              | Print text using green color                                                      |
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -610,31 +610,36 @@ func (o *options) generateTemplate() (*template.Template, error) {
 		"colorMagenta": color.MagentaString,
 		"colorCyan":    color.CyanString,
 		"colorWhite":   color.WhiteString,
-		"levelColor": func(level string) string {
-			var levelColor *color.Color
-			switch strings.ToLower(level) {
-			case "debug":
-				levelColor = color.New(color.FgMagenta)
-			case "info":
-				levelColor = color.New(color.FgBlue)
-			case "warn":
-				levelColor = color.New(color.FgYellow)
-			case "warning":
-				levelColor = color.New(color.FgYellow)
-			case "error":
-				levelColor = color.New(color.FgRed)
-			case "dpanic":
-				levelColor = color.New(color.FgRed)
-			case "panic":
-				levelColor = color.New(color.FgRed)
-			case "fatal":
-				levelColor = color.New(color.FgCyan)
-			case "critical":
-				levelColor = color.New(color.FgCyan)
+		"levelColor": func(value any) string {
+			switch level := value.(type) {
+			case string:
+				var levelColor *color.Color
+				switch strings.ToLower(level) {
+				case "debug":
+					levelColor = color.New(color.FgMagenta)
+				case "info":
+					levelColor = color.New(color.FgBlue)
+				case "warn":
+					levelColor = color.New(color.FgYellow)
+				case "warning":
+					levelColor = color.New(color.FgYellow)
+				case "error":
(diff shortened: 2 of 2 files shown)


tksm:
Hi, thanks for the PR!

I think it would be best to avoid introducing a behavior change for existing users. How about creating a new function, such as `numLevelColor()` or `bunyanLevelColor()`, instead?

hogklint:
Thanks for your reply.

> How about creating a new function, such as `numLevelColor()` or `bunyanLevelColor()`, instead?

If you tail two containers simultaneously, one which uses string levels, and one which uses numeric levels, how could one solve that if there are different color functions with the current behavior of `levelColor`? I haven't been able to catch the error thrown when the type doesn't match so that it could fall back to a numeric level parsing.

If adding a `bunyanLevelColor` and if `levelColor` could be changed to be more lenient this should at least be possible, right? `{{or (levelColor $d.level) (bunyanLevelColor $d.level)}}`. If something similar can be achieved with the current behavior of `levelColor` I would be happy with that, otherwise, I want to suggest this behavior change.

tksm:
> If adding a `bunyanLevelColor` and if `levelColor` could be changed to be more lenient this should at least be possible, right? `{{or (levelColor $d.level) (bunyanLevelColor $d.level)}}`. If something similar can be achieved with the current behavior of `levelColor` I would be happy with that, otherwise, I want to suggest this behavior change.

How about calling `bunyanLevelColor()` before `levelColor()`, instead of changing the current behavior of `levelColor()`? This approach is more flexible because we could add another level color function such as `log4jLevelColor()` in the future.

```bash
for l in 10 20 30 40 50 60 70 \"debug\" \"info\" \"warn\" \"error\" \"fatal\";
  do echo '{"level": '$l', "msg": "A message"}'
done | \
dist/stern --stdin \
  --template='{{with $d := .Message | parseJSON}}[{{or (bunyanLevelColor $d.level) (levelColor $d.level)}}]  {{$d.msg}}{{end}}{{"\n"}}'
```

image

```go
		"bunyanLevelColor": func(level any) string {
			var lv int64
			var err error

			switch value := level.(type) {
			// tryParseJSON yields json.Number
			case json.Number:
				lv, err = value.Int64()
				if err != nil {
					return ""
				}
			// parseJSON yields float64
			case float64:
				lv = int64(value)
			default:
				return ""
			}

			switch {
			case lv < 30:
				return color.New(color.FgMagenta).SprintFunc()(lv)
			case lv < 40:
				return color.New(color.FgBlue).SprintFunc()(lv)
			case lv < 50:
				return color.New(color.FgYellow).SprintFunc()(lv)
			case lv < 60:
				return color.New(color.FgRed).SprintFunc()(lv)
			case lv < 100:
				return color.New(color.FgCyan).SprintFunc()(lv)
			default:
				return strconv.FormatInt(lv, 10)
			}
		},
```

hogklint:
Yeah, sure we can go with that. I just want to point out a couple of things.
* I'm not sure I understand the reason for keeping the current `levelColor` behavior. The error causes the log line to be discarded which seems undesireable to me.
* Having `bunyanLevelColor` fall back to an empty string creates a behavior difference between it and `levelColor` although as a user I would expect them to function similarly.
  * E.g. this `{{or (levelColor $d.level) (bunyanLevelColor $d.level)}}` and this `{{or (bunyanLevelColor $d.level) (levelColor $d.level)}}`  will behave differently even though I would expect the outcome to be the same. Feels like arcane knowledge (which I can try to add to the README).

I'll make the changes, but do let me know if you change your mind. :+1:

hogklint:
Ok, cool. :+1: Even adding the error return value doesn't affect the user. The template engine swallows it and prints it. But I'll stick with the single string return.

hogklint:
All done, let me know if you want to change anything.

tksm (approved):
LGTM 🚀

PR #322, opened 2024-11-12
WIP: --same-prefix-color: same color pods with same prefix

martinpovolny:
First shot (please, suggest a better approach) at 2 issues that I hit.

1) --same-prefix-color -- use the same color for pods that have the same name prefix such as 'api-1' and 'api-2'

2)  Prevent collisions of color assignment if there are enough colors  "query" and "reader"

I have pods from different stateful sets that I would like to color the same, but different for each SS:

```
kubectl get pods
NAME       READY   STATUS    RESTARTS   AGE
api-0      1/1     Running   0          4h46m
api-1      1/1     Running   0          114m
api-2      1/1     Running   0          4h49m
ingest-0   1/1     Running   0          4h47m
ingest-1   1/1     Running   0          4h49m
ingest-2   1/1     Running   0          4h52m
query-0    1/1     Running   0          4h47m
query-1    1/1     Running   0          4h48m
query-2    1/1     Running   0          160m
reader-0   1/1     Running   0          4h46m
reader-1   1/1     Running   0          4h47m
reader-2   1/1     Running   0          4h49m
worker-0   1/1     Running   0          4h43m
worker-1   1/1     Running   0          4h44m
worker-2   1/1     Running   0          4h46m
writer-0   1/1     Running   0          4h48m
writer-1   1/1     Running   0          160m
writer-2   1/1     Running   0          114m
```

with these changes I can do:

```
/stern.bin "(api|ingest|query|reader|worker|writer)-.*" --same-prefix-color --pod-colors "38;2;255;97;136,38;2;169;220;118,38;2;255;216;102,38;2;120;220;232,38;2;171;157;242" --tail 10
```

and get the desired effect 

Please, advise what changes would make this PR or it's part acceptable for the project.

Thank you!

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -89,6 +89,7 @@ type options struct {
 	diffContainer       bool
 	podColors           []string
 	containerColors     []string
+	samePrefixColor     bool
 
 	client       kubernetes.Interface
 	clientConfig clientcmd.ClientConfig
@@ -322,6 +323,7 @@ func (o *options) sternConfig() (*stern.Config, error) {
 		MaxLogRequests:        maxLogRequests,
 		Stdin:                 o.stdin,
 		DiffContainer:         o.diffContainer,
+		SamePrefixColor:       o.samePrefixColor,
 
 		Out:    o.Out,
 		ErrOut: o.ErrOut,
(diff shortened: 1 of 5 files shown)


tksm:
Hi, thanks for the PR!

> --same-prefix-color -- use the same color for pods that have the same name prefix such as 'api-1' and 'api-2'

This is an interesting feature, but I believe it might be challenging to determine the correct prefix in all cases. In your code, the prefix is defined as the part before the first `-`, but for pods like `kube-proxy-knwd2`, users might expect the prefix to be `kube-proxy` (including the `-`), rather than just `kube`.

If you want to colorize pods by their owner objects, like ReplicaSet or Statefulset, we could use `pod.metadata.generateName` or `pod.metadata.ownerReferences` to group the pods by owner.

```yaml
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2024-10-05T11:28:35Z"
  generateName: kube-proxy- # <-- pod.metadata.generateName
  labels:
    controller-revision-hash: 5976bc5f75
    k8s-app: kube-proxy
    pod-template-generation: "1"
  name: kube-proxy-knwd2
  namespace: kube-system
  ownerReferences: # <-- pod.metadata.ownerReferences
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true
    kind: DaemonSet
    name: kube-proxy
    uid: 3cac7f5b-dae4-4b72-9836-c4276a7a2b22
  resourceVersion: "378"
  uid: 5ccd2aa4-37a8-45c5-bf18-f43483c7cc03
```

Issue #323, opened 2024-11-18
ppextjson is not available in templates


fabn:
**What happened**:

I have my containers (some of them) that already emits JSON output, I'd like to use stern to 

* if `.Message` is JSON, parse it and pretty print it
* else output the raw message

According to the home page the function `ppextjson` should be available in templates but actually is not available

With this simple template 

```
{{- if .Message | tryParseJSON -}}
{{- with $d := .Message | parseJSON -}}
{{ json $d | ppextjson | printf "[parsed] %s\n"  }}
{{- end -}}
{{- else -}}
{{ .Message }}
{{- end -}}
```

I get this error

```
Error: unable to parse template: template: log:3: function "ppextjson" not defined
```

**What you expected to happen**:

To mimic the output of this `jq` behavior
```
stern . --output raw | jq -R '. as $line | try (fromjson) catch $line'
```

**Anything else we need to know?**:

Otherwise a simple CLI flag that tells stern that messages are in JSON and (try to) parse them.

**Environment**:
- stern version (use `stern --version`): 1.31.0
- OS (e.g: `cat /etc/os-release`): Sonoma 14.5 (23F79)
- Install tools (e.g: Homebrew): homebrew

tksm:
Thanks for reporting this issue!

It seems that the `README.md` mistakenly mentions that `ppextjson` and `extjson` are available as template functions, when in fact they are only available as `--output` options. These are [pre-defined templates](https://github.com/stern/stern/blob/db6927699d1252bc109b4c48510a68465d84d7d3/cmd/cmd.go#L504-L515) and, unfortunately, are not versatile enough to be used as general-purpose template functions.

fabn:
I understand, to question becomes "is it possible to have a `prettyPrintJSON` function available in templates?`, it should be pretty easy to achieve?

tksm:
I believe formatting the JSON with indentation is easy since we can use `json.MarshalIndent`. However, if you'd like to add syntax highlighting as well, you would need an additional library. PRs are welcome!

fabn:
@tksm I sketched a PR for `prettyJSON` helper

PR #325, opened 2024-11-20
Fix descriptions of `extjson` and `ppextjson`

tksm:
https://github.com/stern/stern/issues/323

This PR updates the descriptions of `extjson` and `ppextjson` and moves them to the output table in `README.md`. Although there is a template function with the same name (`extjson`), it is not versatile enough for general-purpose use, so I propose removing its description from the table of template functions to avoid confusion.


superbrothers (approved):
👍

PR #331, opened 2025-01-06
Add check for when `--no-follow` is set with `--tail=0`

felipecrs:
These flags shouldn't be used together, otherwise no logs will ever be shown.


superbrothers (commented):
Please add a test case.

felipecrs:
> Please add a test case.

Done.

superbrothers (approved):
Thank you for your contribution!

Issue #338, opened 2025-03-13
Highlighting brokes templates


TeaDove:
**What happened**:
Using highlighting (`-i` or `-H`) brokes json templating, because string with highlights is no longer a valid json string

**What you expected to happen**:
Highlighting is applied only after templating done

**How to reproduce it (as minimally and precisely as possible)**:
- Use any template with json parsing, i.e. `-o=ppextjson` and highlighting (`-i` or `-H`)
`stern -o=extjson -s 10m "my-pod" -H "request.processed"`

<img width="1511" alt="Image" src="https://github.com/user-attachments/assets/17fc181a-5a7a-4f44-830a-4ef5622a7758" />

- Templating json and highlighting will fail
Without highlighting everything works: 

<img width="1511" alt="Image" src="https://github.com/user-attachments/assets/3ec7535b-c25e-403f-a162-d207c82be1b4" />
Expected result: 

<img width="1511" alt="Image" src="https://github.com/user-attachments/assets/490a8ae6-4e83-4f12-b2a6-2e20c0565c52" />

**Environment**:
- version: dev 
- OS: macOS 14.3.1 23D60 arm64
- Source (`go install`)

PR #339, opened 2025-03-13
338: fix: move highlighting after formatting

TeaDove:
Closes issue: [338](https://github.com/stern/stern/issues/338)

Highlighting was moved after templating

### DIFF
--- a/stern/file_tail.go
+++ b/stern/file_tail.go
@@ -57,7 +57,7 @@ func (t *FileTail) ConsumeReader(reader *bufio.Reader) error {
 }
 
 // Print prints a color coded log message
-func (t *FileTail) Print(msg string) {
+func (t *FileTail) Print(msg string, withHighlight bool) {
 	vm := Log{
 		Message:        msg,
 		NodeName:       "",
@@ -74,6 +74,10 @@ func (t *FileTail) Print(msg string) {
 		return
 	}
 
+	if withHighlight {
+		fmt.Fprint(t.out, t.Options.HighlightMatchedString(buf.String()))
+		return
+	}
 	fmt.Fprint(t.out, buf.String())
 }
 
@@ -84,6 +88,5 @@ func (t *FileTail) consumeLine(line string) {
 		return
 	}
 
-	msg := t.Options.HighlightMatchedString(content)
-	t.Print(msg)
+	t.Print(content, true)
 }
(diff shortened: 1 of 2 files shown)


TeaDove:
more details here in original issue: https://github.com/stern/stern/issues/338

superbrothers:
I will check the PR this weekend.

stern/tail.go
```diff
@@ -218,6 +218,10 @@ func (t *Tail) Print(msg string) {
 		return
 	}
 
+	if withHighlight {
+		fmt.Fprint(t.out, t.Options.HighlightMatchedString(buf.String()))
+		return
+	}
 	fmt.Fprint(t.out, buf.String())
```
superbrothers, line 225:
```suggestion
	fmt.Fprint(t.out, t.Options.HighlightMatchedString(buf.String()))
```

Some error messages may be highlighted, but I think the above change is fine.

superbrothers (commented):
If possible, I would like you to add tests.

TeaDove:
> If possible, I would like you to add tests.

Sorry for late update, had no time:( 

I've added tests!

stern/tail.go
TeaDove: It's rare case, imho, but if it required, I can add feature to exclude `highlight` from errors

stern/file_tail.go
```diff
@@ -57,7 +57,7 @@ func (t *FileTail) ConsumeReader(reader *bufio.Reader) error {
 }
 
 // Print prints a color coded log message
-func (t *FileTail) Print(msg string) {
+func (t *FileTail) Print(msg string, withHighlight bool) {
```
superbrothers, line 60: Sorry for the late review.

Instead of adding the `withHighlight` argument, I would like to add a new `PrintWithoutHighlight` function that print message without highlight.

```go
func (t *Tail) sprint(msg string) string {
	...
}

// Print prints a color coded log message with the pod and container names
func (t *Tail) Print(msg string) {
	fmt.Fprint(t.out, t.Options.HighlightMatchedString(t.sprint(msg)))
}

// PrintWithNoHighlight prints a log message without applying any highlight.
func (t *Tail) PrintWithoutHighlight(msg string) {
	fmt.Fprint(t.out, t.sprint(msg))
}
```

TeaDove: Done!

superbrothers (approved):

PR #344, opened 2025-03-22
Add support for labels & annotations in templates

### LINKED ISSUE #335: use --template with labels
Use label in template. Like
```
stern -l part-of=microapp --tail 10 --template '{{.ContainerName}} {{.PodName}} {{.Labels.app}} {{.Labels.version}} {{.Message}}'
```
**Why is this needed**:
label is helpful than ContainerName or PodName

nhurden:
Fixes #335 

This is two commits, the first is the minimum needed to support labels & annotations in templates. The second avoids the re-lookup of a pod to get annotations and labels by passing around `Pod` inside `Target`. This should also make it easier to add more fields in templates later as the entire pod is available when `Log` is created.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,13 +147,15 @@ It accepts a custom template through the `--template` flag, which will be
 compiled to a Go template and then used for every log message. This Go template
 will receive the following struct:
 
-| property        | type   | description                                 |
-|-----------------|--------|---------------------------------------------|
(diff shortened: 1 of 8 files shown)


superbrothers (approved):
LGTM

superbrothers:
Thank you for your contribution!

PR #362, opened 2025-12-09
fix: don't stop following logs from pods when receiving EOF

### LINKED ISSUE #361: Log tailing stops after a few hours for long-running pods
**What happened**:

We're (ab)using stern to collect logs from pods inside a Kubernetes namespace when we don't have cluster-wide privileges, nor access to mounting the host filesystem.

In this setup, stern is running for long durations (multiple hours or days). We observed that when tailing logs for long-running pods (again, a pod running for multiple hours), stern eventually stop producing logs for these pods. stern doesn't output any error message.

There are no issues with newly created pods. Restarting stern fixes the issue.

stern keeps tailing logs for all pods, regardless on how long stern or the pods have been running.

I'm able to reproduce this against a GKE Autopilot cluster, directly from my laptop. This was reproduced against multiple Kubernetes versions (at least 1.32 and 1.33, but I suspect the Kubernetes version doesn't matter).

I start a pod in the cluster producing logs (anything will do), then run

```
stern -n stern-reproduction -o json --tail 1 '.*'
```

After 5 or 6 hours, stern will stop outputting logs.

I already have a fix ready, I'm going to open a PR right after this issue. I'm just following the process outlined in the CONTRIBUTING file 😄 

Possibly related: https://github.com/stern/stern/issues/311, although I didn't test if log rotation was the root cause. 

**Environment**:
- stern version (use `stern --version`): 1.33.0
- OS (e.g: `cat /etc/os-release`): MacOS Tahoe for the reproduction, but we encountered this on multiple OS (incl. containers)
- Install tools (e.g: Homebrew): Homebrew for the reproduction, but same issue with directly downloading a Github release
- Others: N/A

Neki:
## What this does

Fix #361

This change updates the logic to decide when the goroutine reading the logs for a pod should exit.

Before this change, receiving an EOF on the API request that fetches the logs would exit the goroutine. It was assumed that an EOF always means "the pod no longer exists". The goroutine is never recreated.

However, I suspect that, at least on GKE clusters, the client may receive an EOF (server gracefully closes the connection) in other situations (maybe a load balancer somewhere has a limit on the duration of an open connection). This would explain why stern eventually stops producing logs for long running pods (see linked issue).

When dumping the goroutine stack traces of a stern process that no longer produced logs, the goroutine that reads logs was missing. There was no error in stern output.

After this change, upon receiving an EOF, we also test whether we received a deletion event from the API (from the goroutine that watches pod creations and deletions). If not, we retry reading the logs.

This assumes that we never miss a pod deletion event, but avoids introducing an extra API request to check whether the pod still exists. I think that should work in practice. 

## How this was tested

I ran this modified version against a cluster that was exhibiting the issue described in #361, and I'm no longer able to reproduce it (whereas before I could reproduce 100% of the time).

## Caveat

I may have misunderstood the issue, or my fix may not be the correct solution. I'm less confident than what the above may seem to convey 🙂

### DIFF
--- a/stern/stern.go
+++ b/stern/stern.go
@@ -17,6 +17,7 @@ package stern
 import (
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"regexp"
 	"strings"
@@ -156,6 +157,16 @@ func Run(ctx context.Context, client kubernetes.Interface, config *Config) error
 			tail.Close()
 			if err == nil {
 				return
+			}
+			if err == io.EOF {
+				// In most cases, EOF signals that the container has terminated.
+				// (and is no longer producting logs).
+				// However, long-running connections might also be closed gracefully by the API server
+				// (or a load balancer in between). Thus, we check whether the target is still active
+				// before stopping tailing logs.
+				if filter.isActive(target) {
+					continue
+				}
 			}
 			if !filter.isActive(target) {
 				fmt.Fprintf(config.ErrOut, "failed to tail: %v\n", err)
--- a/stern/stern.go
+++ b/stern/stern.go
@@ -171,7 +171,11 @@ func Run(ctx context.Context, client kubernetes.Interface, config *Config) error
 				// To ensure we retry fetching the pod logs in that case, we wait a bit before exiting.
 				// If we receive the "pod deleted" event during this time, then the
 				// target will be removed from the filter (see next check) and we won't retry.
-				time.Sleep(2 * time.Second)
+				select {
+				case <-time.After(2 * time.Second):
+				case <-ctx.Done():
+					return
+				}
 			}
 			if !filter.isActive(target) {
 				fmt.Fprintf(config.ErrOut, "failed to tail: %v\n", err)
--- a/stern/stern.go
+++ b/stern/stern.go
@@ -31,6 +31,7 @@ import (
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/time/rate"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/utils/ptr"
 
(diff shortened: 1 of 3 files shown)


Neki:
After further testing, there's an issue with this patch: the error message `failed to tail: pods ... not found` is frequently displayed when a pod exits. It's probably a race condition, where we retry fetching the pod logs before we cancel the goroutine.

Edit: I added a commit to address that, please let me know what you think.

jqywmgycdn-sys (commented):
Hey

tksm:
Hi, thank you very much for the PR!

While testing this change, I noticed a possible difference in behavior when a pod restarts. I might be misunderstanding something, but I wanted to share what I observed along with a simple reproduction case.

The following command prints the current date three times and then exits with exit code 1, which causes the pod to restart:

```bash
kubectl run app --image busybox -- sh -c 'for _ in $(seq 3); do date; sleep 1 ;done; false'
```

With this setup, the pod restarts after emitting three log entries.

In the screenshot below, the left side shows the behavior with the original version of stern, and the right side shows the behavior with the version built from this PR.

Before this change, stern seemed to handle pod restarts as expected. With this change applied, however, I occasionally see the same log entries being displayed multiple times after the pod restarts.

stern

Neki:
Thanks for testing this and the reproduction, it's caused by a new tailer goroutine being spawned while the previous one is still running (and waiting to check whether the pod actually exited or not).
I'll see if I can implement a fix, but I'm worried that the logic is getting convoluted.

Neki:
Hello, so as you can guess from my long silence, I made little progress.

I experimented with different approaches, but they all resulted either in a degradation in the base case (slower detection of containers that exit), being unreliable (such as missing the scenario where the pod is deleted then restarted quickly after), or requiring extra API calls.

The challenge is:

EOF is not a reliable indicator of "pod no longer exists", but the current codebase assumes it is, so `stern` may stop tailing logs for some long-running pods.

However, in most cases, EOF does mean "pod no longer exists". I haven't found a more reliable signal for that:

* watching Kubernetes events didn't work on my test GKE cluster: events could be delayed for several seconds, and in some cases I haven't completely explored, some events weren't delivered at all.
* making an API call to check if the container is still running after getting an EOF results in, on average, one extra API call per container. I think this would be an unwelcome regression to fix an edge case, given that `stern` already implements several controls to limit API calls.

I also tried to refactor the code to properly track the status of each container based on the Kubernetes event streams, rather than the somewhat ad-hoc approach currently used. I stopped because this started touching a lot of code and made the overall code more complex. I don't think this complexity is worth it: `stern` is primarly a dev tool and it doesn't drop logs in 99% of cases.

My conclusion is rather I should stop abusing `stern` as a log shipper and rely on something that's built to not drop logs. Some references for future readers:

* vector has an open issue to add this feature: https://github.com/vectordotdev/vector/issues/23597
* Grafana Alloy has a source that read logs from the Loki API: https://grafana.com/docs/alloy/latest/reference/components/loki/loki.source.kubernetes/ (at work, we had several operational issues with Alloy in production though, so it wouldn't be my favorite option)

PR #367, opened 2026-05-02
fix: close config file after reading

Spuxy:
## Summary
  - The config file opened in `readConfigFromFile` was never closed, leaking the file handle until process exit.
  - Added `defer configFile.Close()` immediately after the successful `os.Open`.

  ## Simple testing
  - [x] `go build ./...`
  - [x] `make build`


tksm (approved):
LGTM. Thanks for the fix!

PR #372, opened 2026-06-18
Keep .Message raw with --timestamps; expose timestamp as .Timestamp

### LINKED ISSUE #342: extractJSONParts in template fails when using --timestamps=short
**What happened**:
```
expanding template failed: template: log:1:32: executing "log" at <extractJSONParts .Message "message">: error calling extractJSONParts: invalid character '3' after top-level value
```

stern to properly show the extracted message field from the JSON log record

use any JSON log record (e.g. from `fluentbit`) and the following command:
`stern -c fluentbit --timestamps=short --template '{{color .PodColor .PodName}}: {{extractJSONParts .Message "message"}}'`

Maybe this should be a separate bug request, but when using the `--template` inline in the CLI we have to add a newline character in the template for stern to output each log entry on a new line, otherwise it concatenates records.

**Environment**:
- stern version (use `stern --version`): `version: 1.32.0`
- OS (e.g: `cat /etc/os-release`): OSX 15.3.2
- Install tools (e.g: Homebrew): Homebrew
- Others:

s3onghyun:
Fixes #342.

### Problem

With `--timestamps`, stern prepends the timestamp onto the log content that becomes `.Message`. So template helpers that parse `.Message` as JSON — `extractJSONParts`, `parseJSON`, `tryParseJSON`, `prettyJSON`, `extjson` — receive `2023-... {"...":...}` instead of `{"...":...}` and fail with `invalid character '<digit>' after top-level value` (exactly the error in #342). The README documents `.Message` as "the log message itself", so this is a contract violation, and there was no separate field to get the timestamp from.

### Fix

Keep `.Message` raw and expose the timestamp as a new `.Timestamp` template field:
- add `Timestamp` to the `Log` struct; `consumeLine` no longer mutates the content, it passes the timestamp through to `sprint`.
- the built-in `default` and `raw` templates render `{{if .Timestamp}}{{.Timestamp}} {{end}}{{.Message}}` — visible output is unchanged when `--timestamps` is on, and byte-identical when off (empty field).
- documented `.Timestamp` in the README.

### Testing

Added `TestConsumeStreamTailTimestamp`: with `--timestamps`, `{{.Message}}` stays raw JSON and `{{.Timestamp}}` carries the timestamp. Fails before the fix (`.Message` polluted; `.Timestamp` undefined), passes after. `go test ./...` green.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,7 @@ will receive the following struct:
 | property        | type              | description                                 |
 |-----------------|-------------------|---------------------------------------------|
 | `Message`       | string            | The log message itself                      |
+| `Timestamp`     | string            | The log timestamp formatted per `--timestamps`/`--timezone`, empty unless `--timestamps` is set |
 | `NodeName`      | string            | The node name where the pod is scheduled on |
 | `Namespace`     | string            | The namespace of the pod                    |
 | `PodName`       | string            | The name of the pod                         |
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -541,12 +541,12 @@ func (o *options) generateTemplate() (*template.Template, error) {
 	if t == "" {
 		switch o.output {
 		case "default":
-			t = "{{color .PodColor .PodName}} {{color .ContainerColor .ContainerName}} {{.Message}}"
+			t = "{{color .PodColor .PodName}} {{color .ContainerColor .ContainerName}} {{if .Timestamp}}{{.Timestamp}} {{end}}{{.Message}}"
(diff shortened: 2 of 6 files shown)


tksm (approved):
LGTM 🥇 Could you rebase onto the latest master?

s3onghyun:
Thanks for the review! Rebased onto the latest master and force-pushed. Tests still pass (`go test ./...`).

tksm:
Thanks for your contribution!

PR #374, opened 2026-07-24
fix: build logs not supported - would be great to have it though

### LINKED ISSUE #358: build logs not supported - would be great to have it though

```
stern build/llm-d-cuda-6
Error: resource type build is not supported
```

**Why is this needed**:
build logs can be long and often timeout when using kubectl or oc log

yhay81:
Fixes #358.

## Summary

build logs not supported - would be great to have it though

## Validation

- Mechanical gate: +41/-1, tests passed
- Adversarial review: approved

---
🤖 AI-authored PR, operated by @yhay81.


tksm:
@yhay81 Thanks for the PR! The code looks good to me. 🚀 

The label `openshift.io/build.name` matches what's defined in [`openshift/api`](https://github.com/openshift/api/blob/18550f1a6d13692eee627b2cad0880c15de07741/build/v1/consts.go#L50-L53), so it looks correct. Could you confirm this was tested on an actual OpenShift cluster? I don't have an OpenShift cluster to verify it myself.

yhay81:
Thanks, and sorry for the delayed response. I have not tested this on an actual OpenShift cluster. The validation so far is limited to the unit tests in this PR and confirming the `openshift.io/build.name` label against `openshift/api`, so real-cluster validation is still pending. I’m happy to keep the PR open for someone with an OpenShift environment to verify it.

tksm:
Looks good! Let's keep this open until someone with an OpenShift environment can verify.

PR #375, opened 2026-07-25
Assign distinct colors to pods created by the same workload

### LINKED ISSUE #347: Anti-affinity for color
When I am watching a `Deployment` with two replicas along with some other pods in a namespace, and the replicas are printing similar messages at similar times, it is particularly important that they be given different colors. Yet the color assignment for new pods seems to be random, regardless of the pod’s roles. While the terminal is limited in how many colors it can render, it would be nice if Stern specifically tried to choose distinct colors for pods that are part of the same `ReplicaSet`, `Deployment`, or `StatefulSet`, so that interleaved output would clearly show which replica was printing which line.

jojinkb:
Fixes #347

## What this PR does

Pod colors are currently chosen by hashing the pod name (`fnv32(podName) % len(colorList)`).
Replicas of the same Deployment/ReplicaSet/StatefulSet differ only in a random suffix, so
each replica is effectively an independent random pick from the color list — with the
default 6 colors, two replicas share a color ~17% of the time and three replicas ~45%.
When replicas print similar messages at similar times, that makes the interleaved output
hard to attribute (#347).

This PR makes stern deliberately spread colors across pods of the same workload:

- Pods are grouped by `metadata.generateName`, which all workload controllers (ReplicaSet,
  StatefulSet, DaemonSet, Job, ...) set, as suggested in
  https://github.com/stern/stern/pull/322#issuecomment-2473257124. The `pod-template-hash`
  is stripped from it so that pods from different ReplicaSets of the same Deployment share a
  group — during a rolling update, old and new pods also get distinct colors.
- A new pod starts at its usual hash-based index and takes the color assigned **least
  recently** within its group. This guarantees distinct colors for pods of the same workload
  as long as unused colors remain in the color list; once the list is exhausted, the least
  recently assigned color (whose previous owner is the most likely to be gone) is reused.
- The assignment is remembered per pod, so a pod keeps its color when its log stream is
  retried/resumed or its container restarts.

## Behavior notes

- Pods that do not share a group with other pods (bare pods, single replicas, mirror pods)
  keep **exactly** the same color as before — the picker falls through to the current
  hash-based index for the first pod of every group.
- For multi-pod groups, a pod's color now depends on the order in which pods are seen
  (the initial list is name-ordered, so it is stable for a static set of pods), instead of
  being a pure function of the pod name — that is inherent to any anti-affinity scheme,
  since distinctness requires knowing which colors are already taken.
- Container colors and `--diff-container` behavior are unchanged, as are all exported APIs
  (`NewTail` signature is untouched).

I implemented this as default behavior (no new flag) since the collision-prone assignment
had no property worth preserving for randomly-named replicas, and singleton pods keep
byte-identical colors. If you would rather gate it behind an opt-in flag like
`--diff-container` (#305), I am happy to add one — the picker is isolated, so it is a small
change.

## How to validate

Unit tests cover the new behavior (`stern/color_test.go`):

- 6 replicas of one ReplicaSet get 6 distinct colors (the chosen names collide under the
  old hash assignment, so this test fails without the fix)
- pods of two ReplicaSets of the same Deployment get distinct colors
- the least recently used color is reused when the list is exhausted
- pods without `generateName` keep the hash-based color
- group-key derivation (Deployment / bare ReplicaSet / StatefulSet / plain pod)

```
go test ./stern/...
```

On a cluster:

```
kubectl create deployment backend --image=nginx --replicas=6
stern deploy/backend
```

Before: replicas regularly share colors (5 or fewer distinct colors for 6 replicas is the
common case). After: all 6 replicas have distinct colors, and during
`kubectl rollout restart deployment backend` the new pods pick colors distinct from the
old ones.

`go build ./...`, `go test ./stern/...`, `go vet ./stern/...`, and
`golangci-lint run ./stern/...` (v2.12.x, repo config) all pass.


tksm (approved):
LGTM, thanks! 🎉  Default behavior makes sense here.
Please rebase onto the latest master.

PR #376, opened 2026-07-25
Add --pager flag to pipe output into a pager

### LINKED ISSUE #345: Better UX of using stern with a pager

It would be nice to have a configuration parameter that makes stern pipe its output to a pager (e.g. `less`).

**Why is this needed**:

Oftentimes, I find myself trying to search the log output with `less`. So if I do just `stern <arguments> | less`, I lose colors in the output. In order to preserve the colors, I need to do something like `stern <arguments> --color=always | less -R`, which is a bit too much typing.

Similar to say [bat](https://github.com/sharkdp/bat?tab=readme-ov-file#automatic-paging), there could be a `--pager` option, which users could put into their configuration file.

jojinkb:
## What this PR does

Adds an opt-in `--pager` flag that makes stern pipe its output into a pager command (e.g. `less`), as requested in #345.

```
stern --pager less backend
```

Since stern spawns the pager itself and stays attached to the terminal, colors are preserved with the default `--color=auto` — no more typing `stern <args> --color=always | less -R`. Like every other flag, it can be set permanently in the config file:

```yaml
pager: less
```

Closes #345

## Behavior details

- **Only used when stdout is a terminal** (same gating as `bat` and `gh`), so redirects and scripts are unaffected even when `pager` is set in the config file.
- **Pager-friendly environment:** the pager runs with `LESS=FRX` and `LV=-c` unless those variables are already set (same defaults as git), so plain `less` shows raw ANSI colors, quits if the output fits on one screen, and doesn't clear the screen on exit.
- **Quitting the pager stops stern:** when the pager exits (e.g. the user presses `q` in `less`), the root context is canceled so stern stops tailing and exits cleanly instead of writing into a broken pipe. `Tail.Start` already treats `context.Canceled` as a normal stop, so no changes to the `stern` package were needed.
- **`--no-follow` keeps the pager open:** once all logs have been shown, stern closes the pipe (EOF) and waits for the pager, so you can keep browsing until you quit.
- The command is split with `github.com/kballard/go-shellquote` (promoted from an indirect to a direct dependency; `go.sum` unchanged), so `--pager 'less -S'` works.
- The pager is started after the interactive `--prompt` handler so the two don't conflict, and stderr (the `+`/`-` tail markers and errors) intentionally bypasses the pager, matching `stern | less` semantics.
- The flag is strictly opt-in: the default (empty) leaves current behavior untouched. If you'd prefer a `$PAGER` environment fallback or a `bat`-style `--paging=auto|always|never` mode, I'm happy to extend it — kept the first iteration minimal on purpose.

## How to validate

Unit tests cover the pager lifecycle (output piped through, context canceled on pager exit, error cases, `LESS`/`LV` environment handling, tty detection):

```
go test ./cmd/...
```

Interactively, with any cluster:

```
# colors preserved inside less; press q and stern exits
stern --pager less <query>

# stern waits for you to quit the pager after showing all logs
stern --no-follow --pager less <query>
```

Without a cluster (pager engages only on a terminal; `--stdin` mode uses the same output path):

```
printf 'l1\nl2\n' | stern --stdin --only-log-lines --pager 'less'
```

`README.md` was regenerated with `make update-readme`, and a "Using a pager" section documents the feature.


tksm (approved):
Nice work! 🎉 Love the pager-friendly environment defaults. Please rebase after #375 is merged.

PR #378, opened 2026-08-09
docs: list toUTC in README template functions

locker95:
`toUTC` is already in the template FuncMap but never made it into the README table. Added it next to the other timestamp helpers.


tksm (approved):
LGTM :+1: Thanks for the PR!

PR #380, opened 2026-08-15
fix: honor --timestamps when reading from stdin

### LINKED ISSUE #379: --timestamps is ignored when reading from --stdin
**What happened**:

With `--stdin`, `--timestamps` is silently ignored. The flag is accepted, and the timestamp is simply absent from the output.

A timestamp on each line, the way `--timestamps` behaves for every other input source.

```
$ printf 'hello from stdin\n' | stern --stdin --timestamps
  hello from stdin
```

Expected something like:

```
  2026-08-15T19:37:17.638305189Z hello from stdin
```

`--timestamps=short` and `--timezone` are ignored in the same way, since they all flow through the same path.

`FileTail.Print` takes no timestamp argument, and `consumeLine` calls it with `""` unconditionally, so `TailOptions.Timestamps` never has an effect on the stdin path. `Tail.Print` in `tail.go` takes the timestamp from the k8s API log line, which stdin lines do not carry, so the stdin path needs its own source for it.

I have a fix and will open a PR referencing this issue.

**Environment**:
- stern version (use `stern --version`): built from master at `3dd605b`, `version: dev`
- OS (e.g: `cat /etc/os-release`): Ubuntu 24.04.4 LTS
- Install tools (e.g: Homebrew): built from source
- Others: go1.26.2

VXNCXNX:
Fixes #379

`--timestamps` is accepted alongside `--stdin` and then silently does nothing.

```
before: $ printf 'hello from stdin\n' | stern --stdin --timestamps
          hello from stdin

after:  $ printf 'hello from stdin\n' | stern --stdin --timestamps
          2026-08-15T19:37:17.638305189Z hello from stdin
```

`FileTail.Print` took no timestamp argument, so `consumeLine` passed `""` unconditionally and `TailOptions.Timestamps` never reached the output. `Tail.Print` in `tail.go` gets its timestamp from the k8s API log line; stdin lines carry no timestamp of their own, so the stdin path needs its own source, and the current time at the moment the line is read is the only one available.

## The change

`FileTail.Print` gains a `timestamp` parameter, matching `Tail.Print`'s signature, and `consumeLine` fills it from `time.Now().UTC()` when `Options.Timestamps` is set. The value goes through `UpdateTimezoneAndFormat`, the same helper `tail.go` uses, so `--timestamps=short` and `--timezone` are handled by the existing code rather than by anything new here. A formatting error is surfaced inline the way `tail.go` surfaces it, rather than dropping the line.

Without `--timestamps` the output is unchanged, and the template variable stays empty as before.

## How to validate

```
printf 'a\n' | stern --stdin --timestamps
printf 'a\n' | stern --stdin --timestamps=short
printf 'a\n' | stern --stdin --timestamps --timezone America/New_York
printf 'a\n' | stern --stdin
```

I ran all four against binaries built before and after. Output:

```
2026-08-15T19:37:22.820975720Z a
08-15 19:37:26 a
2026-08-15T15:37:22.811304621-04:00 a
a
```

Only the last one matches the old behaviour, which is the point.

`TestConsumeFileTailTimestamps` in `stern/file_tail_test.go` covers both branches. It asserts on the `T` separator rather than on a literal instant, since the timestamp is the current time.

Mutating only the condition, so the timestamp branch never runs, fails it:

```
--- FAIL: TestConsumeFileTailTimestamps/with_timestamps
    file_tail_test.go:83: 0: expected output to contain "T", but got "test line\n"
```

`go build ./...`, `go vet ./...`, `go test ./...`, `gofmt -l` and `golangci-lint run ./...` (0 issues) are all clean.

*Disclosure: written with AI assistance (Claude Code). I reproduced the missing timestamp at the CLI, checked all four flag combinations against before and after binaries, and ran the mutation check myself.*


tksm (commented):
Thanks for the fix! 

Could you share a concrete use case where `--stdin` and`--timestamps` would be helpful? For example, when piping `kubectl logs --timestamps` output, the original timestamps are already in the message, so I wonder if it would make more sense to parse those timestamps rather than use `time.Now()`.

VXNCXNX:
Good question, and testing it changed my mind about the patch.

The case in #379 is stdin with no timestamps of its own: `journalctl -f | stern --stdin`, a tailed log file, output from something that is not kubectl. There `--timestamps` has nothing to read.

But your case is worse than "not helpful", it is wrong. Piping `kubectl logs --timestamps` through it today doubles the timestamp:

```
$ printf '2026-08-20T10:00:00.123456789Z hello\n' | stern --stdin --timestamps
  2026-08-20T07:26:16.877339999Z 2026-08-20T10:00:00.123456789Z hello
```

So parsing an existing timestamp is better, and stern already has the piece: `splitLogLine` in `tail.go`. The catch is it splits on the first space unconditionally, so on a plain line it eats the first word as the timestamp. It needs a `time.Parse(time.RFC3339Nano, ...)` guard, which makes it exact rather than a heuristic:

```
"2026-08-20T10:00:00.123456789Z hello"  -> parses, use it
"hello from stdin"                      -> does not parse, no timestamp in the line
```

That leaves what to do in the second case, and I would rather you picked:

1. Parsed timestamp when present, `time.Now()` otherwise. Keeps #379 fixed.
2. Parsed timestamp when present, nothing otherwise. Never invents a time, but leaves `journalctl -f | stern --stdin` with `--timestamps` doing nothing, which is the reported bug.

I lean to 1. Say which and I will push it with tests for both line shapes.

VXNCXNX:
Rebased onto current `master` just now, since the branch had fallen behind by one commit. The change itself is untouched, the only difference is the upstream commit it now sits on. `go build ./...`, `go test ./...`, `gofmt -l` and `go vet ./...` are all clean locally.

One thing worth flagging: the new CI run came back as `action_required` rather than running, so no check has reported on the branch:

```
$ gh pr checks 380
no checks reported on the '379-stdin-timestamps' branch
```

It needs someone to approve the workflow run. The original run on 15 Aug passed before the branch went stale.

Happy to make any changes, or to close it if the approach is not what you want here.

tksm:
Thanks for rebasing. Option 1 looks good to me 👍 . Parsing the existing timestamp when present and falling back to `time.Now()` fixes #379 without breaking the `kubectl logs --timestamps` case.

VXNCXNX:
Option 1 pushed in b992f47.

`splitLogLineIfTimestamped` is `splitLogLine` with a `time.Parse(time.RFC3339Nano, ...)` guard, so the unconditional split stays where it is safe (API lines always carry a timestamp) and a plain stdin line keeps its first word.

```
$ printf '2026-08-20T10:00:00.123456789Z hello from kubectl\nplain line with no timestamp\n' | stern --stdin --timestamps

before  2026-09-14T12:29:01.402277668Z 2026-08-20T10:00:00.123456789Z hello from kubectl
        2026-09-14T12:29:01.402352785Z plain line with no timestamp

after   2026-08-20T10:00:00.123456789Z hello from kubectl
        2026-09-14T12:29:01.598158861Z plain line with no timestamp
```

With `--timestamps` unset the output is byte-identical to before, including for a line that does carry a timestamp, so nothing is stripped from anyone who is piping timestamps through today and reading them in the message.

Tests cover both shapes as promised: a line with its own timestamp reuses it exactly, a line without one gets a generated timestamp that has to parse as `TimestampFormatDefault`, and both are also asserted with `--timestamps` off. `splitLogLineIfTimestamped` has its own table, including `2026-08-20 10:00:00 hello`, where the first field alone is not a valid timestamp and the line must be left whole.

Four mutations each fail a different case: always using `time.Now()`, reusing the timestamp without stripping it, dropping the parse guard, and removing the fallback. `go test ./...`, `gofmt -l` and `go vet ./...` are clean.

CI is still showing `action_required` on the branch, so it will need a workflow approval to report.

tksm:
The code looks good! Could you rebase onto the latest `master`? Sorry for the extra step.

VXNCXNX:
Rebased onto `523c01a`, no extra step needed on your side. No conflicts, and both commits applied unchanged.

`go build ./...`, `go test ./...`, `gofmt -l` and `go vet ./...` are clean on the new base, and the behaviour is unchanged:

```
2026-08-20T10:00:00.123456789Z hello from kubectl
2026-09-14T12:50:37.193479509Z plain line with no timestamp
```

tksm (approved):
LGTM 👍 Thanks for the PR!

PR #382, opened 2026-08-19
Fix misspelled identifiers reHightlight and stirngSlice

MsfPablo:
Two spelling fixes on internal identifiers:

- `reHightlight` -> `reHighlight` (unexported field in `stern/tail_utils.go`)
- `stirngSlice` -> `stringSlice` (local variable in `hack/update-readme/update-readme.go`)

Both are package-private with all references in the same file, so nothing outside changes. `go build ./...` and `go vet` pass.


tksm (approved):
Thanks for the fix! LGTM 👍

